PR #19, opened 2022-04-18
Update demo to add in notes on depot_tools and tensorflow versioning

boomanaiden154:
Added some misc notes to the documentation regarding tensorflow versioning. When I tried getting my initial compile setup working, I'm highly confident that compiling with Tensorflow v2.8.0 caused the build to fail (Tensorflow v2.7.0 isn't compatible with Python 3.10 on Ubuntu 22.04, so I had to downgrade to 20.04), and then I ran into some linking issues when I tried building with a more recent version of libtensorflow to try and make things compatible. This seems to have been noticed in commit e11c886edd7e1e362fae6bbae4f2fca659b22fb3, and I'm going to try and do some more investigation into this issue to see if I can get the libtensorflow version upgraded. However, the buildbot script seems to be using v2.6.0 happily, so maybe there's just something specific with the configuration of the demo?

### DIFF
--- a/docs/demo/demo.md
+++ b/docs/demo/demo.md
@@ -108,7 +108,20 @@ This is this repository.
 
 ### Tensorflow dependencies
 
-See also [the build bot script](buildbot/builbot_init.sh)
+See also [the build bot script](../../buildbot/buildbot_init.sh)
+
+**NOTE:** The versioning of Tensorflow is pretty important in order to get a
+functioning build with the demo setup. Building with a Tensorflow pip
+package above v2.7.0 will cause the build to fail as well as building with
+a libtensorflow version above v1.15.x. Versions of the Tensorflow pip package
+under v2.8.0 don't have any available wheels for Python 3.10, so if you
+are on a more recent distro (eg Ubuntu 22.04) that has Python 3.10, you will
+run into version compatibility issues between the required Tensorflow version
+and the available Python version (pip will refuse to install Tensorflow) trying
+to install the required python packages from `ml-compiler-opt/requirements.txt`.
+To mitigate this, either use a distro with a compatible python version
+(eg Ubuntu 20.04), or your preferred flavor of python virtual environment
+to utilize a compatible Python version.
 
 ```shell
 cd
--- a/docs/demo/demo.md
+++ b/docs/demo/demo.md
@@ -247,7 +247,7 @@ rm -rf $DEFAULT_TRACE &&
     --data_path=$CORPUS \
     --output_path=$DEFAULT_TRACE \
     --gin_files=compiler_opt/rl/inlining/gin_configs/common.gin \
-    --gin_bindings=config_registry.get_configuration.implementation=@configs.InliningConfig
+    --gin_bindings=config_registry.get_configuration.implementation=@configs.InliningConfig \
     --gin_bindings=clang_path="'$LLVM_INSTALLDIR/bin/clang'" \
     --gin_bindings=llvm_size_path="'$LLVM_INSTALLDIR/bin/llvm-size'" \
     --sampling_rate=0.2


docs/demo/demo.md
```diff
@@ -49,6 +49,24 @@ export LLVM_SRCDIR=~/llvm-project
 export LLVM_INSTALLDIR=~/llvm-install
 ```
 
+### Depot Tools
```
petrhosek, line 52: This shouldn't be needed. Jiri downloads CIPD and installs it in `${FUCHSIA_SRCDIR}/.jiri_root/bin/cipd`. If you follow the instructions at https://fuchsia.dev/fuchsia-src/get-started/get_fuchsia_source#set-up-environment-variables and set the environment variables as suggested, `cipd` should be in your `PATH`.

yundiqian:
@mtrofin Could you take a look at the TF versioning issue?

mtrofin:
> @mtrofin Could you take a look at the TF versioning issue?

That looks like one of the problems that the move to TFLite will solve - like @boomanaiden154 observed, the main action item is that the versions spelled out *really* matter. In principle they kind of always do, but we expect with TFLite to matter less, and also to not have interdeps like TF C API has now with the pip wheel.

boomanaiden154:
@mtrofin Has there been any work done so far on a move to TFLite? Once I have some more time (approximately three weeks), I'd be interested in transitioning everything over to TFLite before GSoC if no one else is already working on it.

mtrofin:
> @mtrofin Has there been any work done so far on a move to TFLite? Once I have some more time (approximately three weeks), I'd be interested in transitioning everything over to TFLite before GSoC if no one else is already working on it.

yes - I have the patch ready; that, turns out, was the easy part (may need a bit of love if it got out of sync tho). @petrhosek is wrapping up the cmake part, which turned out to be the hard part. Meanwhile @yundiqian pointed out it'd probably be best we allowed both TF and TFLite to co-exist, maybe as a compile-time switch; that should be easy to do tho. There's also a bit of buildbot massaging - like we'd want to keep building the TF way. I can help with that (we need to allocate one), but long story short, if this is still in the air when you start, I'm happy to pass it on, thanks for offering!

boomanaiden154:
Sounds good. Has there been any work done internally on behavioral cloning warm start based on the existing models? That would be another (mini?)project that I would be interested in working on before GSoC if no one else is currently working on it and doesn't seem like it would be too difficult based on my preliminary assessment of the code involved.

mtrofin:
> Sounds good. Has there been any work done internally on behavioral cloning warm start based on the existing models? That would be another (mini?)project that I would be interested in working on before GSoC if no one else is currently working on it and doesn't seem like it would be too difficult based on my preliminary assessment of the code involved.

The code here is the same we use internally. BC-ing from an existing model should work - but maybe I didn't understand your idea, but ya, feel free to play around!

boomanaiden154:
@mtrofin Do you mind merging this when you get a chance? If it's not something you want to include in the documentation, let me know so that I can close the PR.

mtrofin:
> @mtrofin Do you mind merging this when you get a chance? If it's not something you want to include in the documentation, let me know so that I can close the PR.

Happy to merge - I was waiting for the pending comment to be addressed  (line 137: "Can you add a comment on what those issues are and how to mitigate them?")

docs/demo/demo.md
boomanaiden154: Thanks for catching that. Should be fixed after the latest force push.

boomanaiden154:
> Happy to merge - I was waiting for the pending comment to be addressed (line 137: "Can you add a comment on what those issues are and how to mitigate them?")

Are you talking about the comment by Petr Hosek? I don't see anything on line 137.

mtrofin (changes requested):

mtrofin:
> > Happy to merge - I was waiting for the pending comment to be addressed (line 137: "Can you add a comment on what those issues are and how to mitigate them?")
> 
> Are you talking about the comment by Petr Hosek? I don't see anything on line 137.

Sorry - I did "start review" and forgot to actually submit the review. Fixed now.

My comment is a nit - basically either say what the issues are - i.e. dependency issues, or what - or rephrase to something more definitive.

docs/demo/demo.md
```diff
@@ -108,7 +126,15 @@ This is this repository.
...
-See also [the build bot script](buildbot/builbot_init.sh)
+See also [the build bot script](../../buildbot/buildbot_init.sh)
+
+**NOTE:** The versioning of Tensorflow is pretty important in order to get a 
+functioning build with the demo setup. Building with a Tensorflow pip
+package above v2.7.0 will cause the build to fail as well as building with
+a libtensorflow version above v1.15.x. Versions of the Tensorflow pip package
+under v2.8.0 don't have any available wheels for Python 3.10, so if you
+are on a more recent distro (eg Ubuntu 22.04), you might run into issues
+trying to install the required python packages from `ml-compiler-opt/requirements.txt`.
```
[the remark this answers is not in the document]
boomanaiden154: Should be fixed in the latest version. I specified the specific type of issue (version incompatibility, pip will refuse to install required Python version), as well as potential mitigations (using a distro with a compatible python version/python virtual environment with a compatible python version).

mtrofin (approved):
Thanks!

PR #23, opened 2022-06-08
Modify regalloc runner to prevent needing a ThinLTO corpus for generating a default trace

boomanaiden154:
Currently, when running `generate_default_trace.py` against the regalloc problem, it errors out when not using a ThinLTO corpus due to the expansion of the `file_paths` variable. This patch preserves functionality for the ThinLTO case while also allowing running `generate_default_trace` against the regalloc runner without using a ThinLTO corpus.

### DIFF
--- a/compiler_opt/rl/regalloc/regalloc_runner.py
+++ b/compiler_opt/rl/regalloc/regalloc_runner.py
@@ -72,7 +72,11 @@ class RegAllocRunner(compilation_runner.CompilationRunner):
     log_path = os.path.join(working_dir, 'log')
     output_native_path = os.path.join(working_dir, 'native')
 
-    input_ir_path, cmd_path, thinlto_index_path = file_paths
+    input_ir_path = cmd_path = thinlto_index_path = None
+    if len(file_paths) == 3:
+        input_ir_path, cmd_path, thinlto_index_path = file_paths
+    else:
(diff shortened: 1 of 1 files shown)


yundiqian:
@mtrofin is it a reasonable use case for training register allocation policy with a non-thinlto corpus?

mtrofin:
> @mtrofin is it a reasonable use case for training register allocation policy with a non-thinlto corpus?

Yes, any optimization should work with either configuration (with/without thinlto). We ultimately want to validate in realistic industrial setups, but there's no reason to force the tools to only work like that; this allows one rapidly prototype and test new work (easier and quicker to extract non-thinlto corpora), and then move to "fully-fledged" training (with thinlto) next.

mtrofin (approved):

yundiqian:
ic, @boomanaiden154 thanks for the fix!

Just for better code purpose, can you add a check in the code that len(file_paths) == 2 or 3 before the if len(file_paths) == 3 else branch, and raise an error if it is not the case? in this way an unexpected error would be caught earlier in a more explicit way

boomanaiden154:
@yundiqian I just opened a PR (#24) to address your concerns. Let me know if you need anything changed up (eg error message, additional checks, etc).

PR #24, opened 2022-06-08
Added assert to ensure length of file_paths is valid

boomanaiden154:
Adds an assert to make sure that the length of `file_paths` in `_compile_fn` in `regalloc_runner.py` is valid. Should theoretically be limited by the declaration of the parameters as a tuple, but this still allows for smaller values, which we also want to detect and throw an error before anything else happens.

### DIFF
--- a/compiler_opt/rl/regalloc/regalloc_runner.py
+++ b/compiler_opt/rl/regalloc/regalloc_runner.py
@@ -20,6 +20,7 @@ import io
 import os
 import tempfile
 from typing import Dict, Optional, Tuple
+from absl import logging
 
 import gin
 import tensorflow as tf
@@ -40,6 +41,7 @@ class RegAllocRunner(compilation_runner.CompilationRunner):
       ir_path, tf_policy_path, default_reward, moving_average_reward)
   """
 
+  # TODO: refactor file_paths parameter to ensure correctness during construction
   def _compile_fn(
       self, file_paths: Tuple[str, str, str], tf_policy_path: str,
       reward_only: bool, cancellation_manager: Optional[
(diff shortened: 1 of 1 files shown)


yundiqian:
raise error is preferred compared to assert here, one reason being that assert check will be disabled with optimizations, so

if len == 3:
 ...
elif len == 2:
...
else:
  raise ValueError("")

mtrofin:
> raise error is preferred compared to assert here, one reason being that assert check will be disabled with optimizations, so
> 
> if len == 3: ... elif len == 2: ... else: raise ValueError("")

how about logging.fatal then - reason being, raising an error presumes the caller can handle it; but it's not like the caller in this case "held their cards" and could give a valid input next time; rather, different sized inputs are a bug, so we should just crash immediately.

A good next step would be to maybe encapsulate this in a specific data structure that's correct by construction, but no need to complicate this change.

compiler_opt/rl/regalloc/regalloc_runner.py
```diff
@@ -74,9 +76,11 @@ def _compile_fn(
 
     input_ir_path = cmd_path = thinlto_index_path = None
     if len(file_paths) == 3:
-        input_ir_path, cmd_path, thinlto_index_path = file_paths
+      input_ir_path, cmd_path, thinlto_index_path = file_paths
+    elif len(file_paths) == 2:
+      input_ir_path, cmd_path = file_paths
     else:
-        input_ir_path, cmd_path = file_paths
+      logging.fatal("Expected 2 or 3 file paths")
```
yundiqian, line 83: nit: could you use '' instead of "" for the consistency of the file, otherwise lgtm

thanks for the fix!

boomanaiden154: I just changed it to `"Instead of 2 or 3 file paths, got unexpected value"`. Let me know if you want me to change it from this.

yundiqian: oh, i just mean use single quote '' instead of double quote "" for the string message cause the other places in this files are all using single quote '' (see line 72 and 73 above).

Sorry for the confusion -_-!

boomanaiden154: That makes so much more sense. Sorry for the confusion. Should be fixed in the latest version.

mtrofin:
On the `logging.fatal` stuff, @yundiqian and @Northbadge convinced me we should move to raising `ValueError`. Apologies for the confusion - wasn't aware of https://peps.python.org/pep-0008/#programming-recommendations (thanks @Northbadge) and, while the idea of raising errors for fast-fail still makes me uncomfortable, I understand that's in a context in Python where there is some design and guidance mindful of it, so we should follow that guidance. 

I'll follow up with a patch (I'll also want to see if I can get pylint to complain if the few places we want to catch-all BaseException have a justification next to them)

PR #26, opened 2022-06-13
Add candidate instruction features for regalloc model

boomanaiden154:
This patch adds in the `lr_use_def_instructions` feature, currently specifically capturing instructions regarding the candidate virtual register. Up to 33 use/def instructions are passed to the model (zero-padded otherwise, 33 so that the concatenate layer will function properly), and then passed to an embedding layer. The output of this embedding layer is then passed through to the concatenate layer which then gets piped into the rest of the model. Associated vocab changes are also present.


boomanaiden154:
Correct PR is [here](https://github.com/boomanaiden154/ml-compiler-opt/pull/1). Patch is not anywhere close to being ready to merge into upstream/main.

PR #28, opened 2022-06-21
Update demo documentation

Northbadge:


boomanaiden154:
It seems like you were able to get this built with libtensorflow 2.x? When I'm following the steps in the demo, I'm still getting a linking error when trying to use libtensorflow 2.x instead of 1.x (even with head of tree LLVM).

Northbadge:
Yep, I managed to build with 2.x. I might have missed out a flag, I’ll check in the morning, thanks for the heads up!

PR #29, opened 2022-06-24
Refactored regalloc gin configs

boomanaiden154:
This should help mitigate issues in the future like the one fixed in #27 as it refactors the gin configs for regalloc so that the common network parameters are stored in their own file and imported into the behavioral cloning and agent configs. Reducing code duplication and having a central place to make changes to the model should allow for easier changes and changes that are more likely to be correct. The same thing could probably also be done with the inlining config although with some differences due to the fact that there are many more configs (multiple common network configs need to be refactored out), and the standard behavioral cloning and agent configs use different network classes. Could be done with gin macros, but it doesn't refactor as nicely as the regalloc configs.

### DIFF
--- a/compiler_opt/rl/regalloc/gin_configs/behavioral_cloning_nn_agent.gin
+++ b/compiler_opt/rl/regalloc/gin_configs/behavioral_cloning_nn_agent.gin
@@ -7,22 +7,14 @@ import tf_agents.agents.behavioral_cloning.behavioral_cloning_agent
 import tf_agents.networks.actor_distribution_network
 
 include 'compiler_opt/rl/regalloc/gin_configs/common.gin'
+include 'compiler_opt/rl/regalloc/gin_configs/network.gin'
 
 train_eval.agent_name=%constant.AgentName.BEHAVIORAL_CLONE
 train_eval.num_iterations=10000
 train_eval.batch_size=64
 train_eval.train_sequence_length=1
 
-regalloc.config.get_observation_processing_layer_creator.quantile_file_dir='compiler_opt/rl/regalloc/vocab'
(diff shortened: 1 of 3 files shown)


compiler_opt/rl/regalloc/gin_configs/behavioral_cloning_nn_agent.gin
```diff
@@ -7,22 +7,16 @@ import tf_agents.agents.behavioral_cloning.behavioral_cloning_agent
 import tf_agents.networks.actor_distribution_network
 
 include 'compiler_opt/rl/regalloc/gin_configs/common.gin'
+include 'compiler_opt/rl/regalloc/gin_configs/network_common.gin'
```
yundiqian, line 10: just call it network.gin as it's essentially defining the network

compiler_opt/rl/regalloc/gin_configs/behavioral_cloning_nn_agent.gin
```diff
@@ -7,22 +7,16 @@ import tf_agents.agents.behavioral_cloning.behavioral_cloning_agent
...
 train_eval.agent_name=%constant.AgentName.BEHAVIORAL_CLONE
 train_eval.num_iterations=10000
 train_eval.batch_size=64
 train_eval.train_sequence_length=1
 
-regalloc.config.get_observation_processing_layer_creator.quantile_file_dir='compiler_opt/rl/regalloc/vocab'
-regalloc.config.get_observation_processing_layer_creator.with_sqrt = False
-regalloc.config.get_observation_processing_layer_creator.with_z_score_normalization = False
-
 create_agent.policy_network = @regalloc_network.RegAllocNetwork
```
yundiqian, line 17: you may also move this to network.gin

PR #30, opened 2022-06-26
Only generate/use vocab when needed

boomanaiden154:
Currently, the vocab generator generates quantile buckets for all the different features regardless of whether or not they are relevant. This patch adds an option into the problem config so that features that don't require quantile buckets won't have them generated. This saves on some performance during vocab generation and prevents having to load a couple things during the preprocessing layer generation (although the latter difference is extremely insignificant). In addition to the config option, this patch also ties the config option into the vocab generator and both current problem configs. The major advantage of this feature is that it skips vocab generation for extremely large features (eg the new instruction based features for the regalloc model) that don't need vocab generated and use up a massive amount of memory (hundreds of gigabytes) during vocab generation for no real reason. I have already validated the changes against both the inlining config (based on the demo), and the regalloc config (based on [these](https://github.com/boomanaiden154/gsoc2022-notes/blob/master/june/3/regalloc-model-training.md) instructions).

### DIFF
--- a/compiler_opt/rl/inlining/config.py
+++ b/compiler_opt/rl/inlining/config.py
@@ -97,3 +97,7 @@ def get_observation_processing_layer_creator(quantile_file_dir=None,
                                      with_z_score_normalization, eps))
 
   return observation_processing_layer
+
+def get_nonnormalized_features():
+  return ['reward', 'inlining_default', 'inlining_decision']
+
--- a/compiler_opt/rl/problem_configuration.py
+++ b/compiler_opt/rl/problem_configuration.py
@@ -93,10 +93,12 @@ class ProblemConfiguration(metaclass=abc.ABCMeta):
       self) -> Callable[[types.TensorSpec], tf.keras.layers.Layer]:
     raise NotImplementedError
 
+  def get_nonnormalized_features(self) -> Iterable[str]:
+    return []
+
   @abc.abstractmethod
   def get_runner_type(self) -> 'type[compilation_runner.CompilationRunner]':
     raise NotImplementedError
-
 
 def is_thinlto(module_paths: Iterable[str]) -> bool:
   return tf.io.gfile.exists(next(iter(module_paths)) + '.thinlto.bc')
--- a/compiler_opt/tools/sparse_bucket_generator.py
+++ b/compiler_opt/tools/sparse_bucket_generator.py
@@ -21,14 +21,17 @@ by exclusively using OSS-compatible libraries.
 import math
 import multiprocessing as mp
 import os
-from typing import Callable, Dict, List
+from typing import Callable, Dict, List, Iterable
 
 from absl import app
 from absl import flags
 from absl import logging
+import gin
 
 import numpy as np
 import tensorflow as tf
+
+from compiler_opt.rl import registry
 
 flags.DEFINE_string('input', None,
                     'Path to input file containing tf record datasets.')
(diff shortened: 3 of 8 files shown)


compiler_opt/rl/problem_configuration.py
```diff
@@ -95,6 +95,10 @@ def get_preprocessing_layer_creator(
   def get_runner(self, *args, **kwargs) -> compilation_runner.CompilationRunner:
     raise NotImplementedError
 
+  @abc.abstractmethod
+  def get_nonnormalized_features(self) -> Iterable[str]:
+    raise NotImplementedError
```
mtrofin, line 100: can this have a default implementation returning []?

boomanaiden154: Fixed. Also removed the `@abc.abstractmethod` decorator so that if it doesn't have to be overridden.

compiler_opt/rl/inlining/config.py
```diff
@@ -97,3 +97,8 @@ def observation_processing_layer(obs_spec):
                                      with_z_score_normalization, eps))
 
   return observation_processing_layer
+
+def get_nonnormalized_features():
+  return ['reward', 'inlining_default',
```
yundiqian, line 102: nit: make them in the same line?

compiler_opt/rl/problem_configuration.py
```diff
@@ -95,6 +95,9 @@ def get_preprocessing_layer_creator(
   def get_runner(self, *args, **kwargs) -> compilation_runner.CompilationRunner:
     raise NotImplementedError
 
+  def get_nonnormalized_features(self) -> Iterable[str]:
```
yundiqian, line 98: nit: move it after get_preprocessing_layer_creator and before runner to be consistent of other files

docs/demo/demo.md
```diff
@@ -262,6 +262,8 @@ in the trace changes.
 rm -rf $DEFAULT_VOCAB &&
   PYTHONPATH=$PYTHONPATH:. python3 \
     compiler_opt/tools/sparse_bucket_generator.py \
+    --gin_files=compiler_opt/rl/inlining/gin_configs/common.gin \
+    --gin_bindings=config_registry.get_configuration.implementation=@configs.InliningConfig \
```
yundiqian, line 266: this is not needed cause it's in the first line of common.gin

compiler_opt/tools/sparse_bucket_generator.py
```diff
@@ -123,17 +135,23 @@ def _generate_vocab(feature_values_arrays, feature_name):
...
 def main(_) -> None:
+  gin.parse_config_files_and_bindings(
+      FLAGS.gin_files, bindings=FLAGS.gin_bindings, skip_unknown=False)
+  logging.info(gin.config_str())
+
   """Generate num_buckets quantiles for each feature."""
   tf.io.gfile.makedirs(FLAGS.output_dir)
   dataset = tf.data.Dataset.list_files(FLAGS.input)
   dataset = tf.data.TFRecordDataset(dataset)
+  problem_config = registry.get_configuration()
```
yundiqian, line 146: nit: move it right after line 140 cause logically they are closer --- we parse the gin in order to get features_to_not_process

docs/demo/demo.md
boomanaiden154: The `--gin_bindings` flag should also probably be removed on a couple other parts of the demo where it is just used to set the config. I can probably open up a PR to fix that, but it isn't a super critical issue.

yundiqian (approved):

Issue #31, opened 2022-06-27
[infra] Support distributed compilation


mtrofin:
This tracks incoming work that aims to support distributing the data collection steps (roughly what `CompilationRunner` implementations do today). The design is mainly aimed at (exclusively) supporting distribution via [dask.distributed](https://distributed.dask.org/en/stable/).

PR #33, opened 2022-06-29
[infra] Introduce a simple local-process worker pool manager

mtrofin:
This is meant as an in-place replacement of the current functionality,
but using a stateful worker object abstraction analogous to what Dask
supports. For local workloads, this implementation is faster than Dask's
LocalCluster (which may be due to misconfiguration). Because it's very
simple and introduces no new dependencies, even if LocalCluster
performance could be improved, the implementation would still be useful
for debugging.

The worker abstraction allows implementers specify a list of methods
that should be executed promptly on the server side, analogous to Dask's
`separate_thread=False` concept. For our purposes, we can use it to
implement full cancelation of work on the server side.

This patch introduces the worker manager, a subsequent patch will enable
its use in the rest of the codebase.

Issue #31

### DIFF
--- a/compiler_opt/distributed/local/local_worker_manager.py
+++ b/compiler_opt/distributed/local/local_worker_manager.py
@@ -39,7 +39,7 @@ import threading
 from compiler_opt.distributed.worker import Worker
 
 from contextlib import AbstractContextManager
-from typing import Any, Optional
+from typing import Any, Dict, Optional
 
 
 @dataclasses.dataclass(frozen=True)
@@ -111,7 +111,7 @@ def _make_stub(cls: 'type[Worker]', *args, **kwargs):
               **kwargs))
       # lock for the msgid -> reply future map
       self._lock = threading.Lock()
-      self._map: dict[int, concurrent.futures.Future] = {}
+      self._map: Dict[int, concurrent.futures.Future] = {}
       # thread drainig the receive queue
       self._pump = threading.Thread(target=self._msg_pump)
 
--- a/compiler_opt/distributed/local/local_worker_manager.py
+++ b/compiler_opt/distributed/local/local_worker_manager.py
@@ -59,7 +59,7 @@ class TaskResult:
 
 
 def _run(in_q: 'queue.Queue[Task]', out_q: 'queue.Queue[TaskResult]',
-         worker_class: type[Worker], *args, **kwargs):
+         worker_class: 'type[Worker]', *args, **kwargs):
   """Worker process entrypoint."""
 
   pool = concurrent.futures.ThreadPoolExecutor()
--- a/compiler_opt/distributed/local/local_worker_manager.py
+++ b/compiler_opt/distributed/local/local_worker_manager.py
@@ -58,7 +58,7 @@ class TaskResult:
   value: Any
 
 
-def _run(in_q: queue.Queue[Task], out_q: queue.Queue[TaskResult],
+def _run(in_q: 'queue.Queue[Task]', out_q: 'queue.Queue[TaskResult]',
          worker_class: type[Worker], *args, **kwargs):
   """Worker process entrypoint."""
 
@@ -96,8 +96,8 @@ def _make_stub(cls: 'type[Worker]', *args, **kwargs):
     """Client stub to a worker hosted by a process."""
 
     def __init__(self):
-      self._send: queue.Queue[Task] = multiprocessing.get_context().Queue()
-      self._receive: queue.Queue[TaskResult] = multiprocessing.get_context(
+      self._send: 'queue.Queue[Task]' = multiprocessing.get_context().Queue()
+      self._receive: 'queue.Queue[TaskResult]' = multiprocessing.get_context(
       ).Queue()
 
       # this is the process hosting one worker instance.
--- a/compiler_opt/distributed/local/local_worker_manager.py
+++ b/compiler_opt/distributed/local/local_worker_manager.py
@@ -1,0 +1,201 @@
+# coding=utf-8
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Local Process Pool - based middleware implementation.
+
+This is a simple implementation of a worker pool, running on the local machine.
+Each worker object is hosted by a separate process. Each worker object may
+handle a number of concurrent requests. The client is given a stub object that
+exposes the same methods as the worker, just that they return Futures.
+
+There is a pair of queues between a stub and its corresponding process/worker.
+One queue is used to place tasks (method calls), the other to receive results.
+Tasks and results are correlated by a monotonically incrementing counter
+maintained by the stub.
+
+The worker process dequeues tasks promptly and either re-enqueues them to a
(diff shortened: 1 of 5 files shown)


compiler_opt/distributed/worker.py
```diff
@@ -0,0 +1,23 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Common abstraction for a worker contract."""
+
+
+class Worker:
+
+  @classmethod
+  def is_priority_method(cls, method_name: str) -> bool:
```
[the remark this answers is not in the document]
mtrofin: Please see the PR description.

compiler_opt/distributed/local/local_worker_manager.py
```diff
@@ -0,0 +1,220 @@
...
+
+
+class LocalWorkerPool(AbstractContextManager):
+  """A pool of workers hosted on the local machines, each in its own process."""
+
+  def __init__(self, worker_class: 'type[Worker]', count: Optional[int], *args,
+               **kwargs):
+    if not count:
+      count = multiprocessing.cpu_count()
+    self._stubs = [
```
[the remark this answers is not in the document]
mtrofin: That happens above this layer. This layer is only concerned with setting up the objects and the communication to them.

compiler_opt/distributed/local/local_worker_manager.py
```diff
@@ -0,0 +1,220 @@
...
+    def _is_cancelled_or_broken(self):
+      return self._stop_pump.is_set() or not self._process.is_alive()
+
+    def _msg_pump(self):
+      while not self._is_cancelled_or_broken():
+        try:
+          # avoid blocking so we may notice if _stop_pump was set. We aren't
+          # very concerned with delay between shutdown request and the message
+          # pump noticing it, this happens only at the very end of training.
+          task_result = self._receive.get(timeout=1.0)
```
Northbadge, line 148: Could avoid the timeout/try/except by feeding a poison pill (e.g. `None`) into the queue and breaking out of the loop if it's detected. Only edge case I can think of is externally killing the child process, not sure if you can feed a poison pill after that

mtrofin: If the external process dies unintentionally (i.e. you got a bug - like illustrated in the test), then the poison pill won't ever come, and the pump will hang out forever.

compiler_opt/distributed/local/local_worker_manager.py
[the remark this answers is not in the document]
yundiqian: will there be another abstract layer above this layer taking care of the load balancing?

yundiqian (approved):

compiler_opt/distributed/local/local_worker_manager.py
Northbadge: adding a `finally:` block which feeds the poison pill to `_run` should do the trick for that

compiler_opt/distributed/local/local_worker_manager.py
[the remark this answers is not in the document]
mtrofin: initially, not an abstract layer, just the way the work will be doled out in local_data_collector.py - roundrobin over the workers.

compiler_opt/distributed/local/local_worker_manager.py
Northbadge: The main shortcoming would be the 1s timeout, which on average would be a 0.5s wait, and I believe we’ll be killing the processes fairly frequently too. On a 10s training cycle that’d be 5%

I think we’d only need to guarantee the correctness of _run outside the try/catch/finally block, which is trivially correct since it’s nothing. (I suppose starting the processes needs to be checked too for completeness) 
_run_impl is free to throw exceptions, since _run will put the poison pill in regardless because of the finally block (except, possibly, that one edge case I noted above)


Also, catching exceptions is [“slow”](https://stackoverflow.com/questions/2522005/cost-of-exception-handlers-in-python)

compiler_opt/distributed/local/local_worker_manager.py
```diff
@@ -0,0 +1,220 @@
...
+    ]
+
+  def __enter__(self):
+    return self._stubs
+
+  def __exit__(self, *args, **kwargs):
+    # each shutdown may take ~1 second because of the timeout in the message
+    # pump, so let's parallelize shutting everything down.
+    with concurrent.futures.ThreadPoolExecutor() as tpe:
+      concurrent.futures.wait([tpe.submit(s.shutdown) for s in self._stubs])
```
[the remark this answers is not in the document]
mtrofin: agreed, addressing.

compiler_opt/distributed/local/local_worker_manager.py
mtrofin: We're not killing these processes ever except for at the very end of training.

mtrofin: ...but - we can just inject `None` at shutdown instead. Doing that.

compiler_opt/distributed/local/local_worker_manager.py
```diff
@@ -0,0 +1,230 @@
...
+
+      return remote_call
+
+    def shutdown(self):
+      try:
+        self._process.kill()
+      finally:
+        pass
+
+    def wait_for_msg_pump_exit(self):
```
Northbadge, line 198: could probably just call this `join()` for consistency, lgtm otherwise :)

compiler_opt/distributed/local/local_worker_manager.py
```diff
@@ -0,0 +1,230 @@
...
+                    is_urgent=cls.is_priority_method(name)))
+            self._map[msgid] = result_future
+        return result_future
+
+      return remote_call
+
+    def shutdown(self):
+      try:
+        self._process.kill()
+      finally:
```
yundiqian, line 195: what's this for?

compiler_opt/distributed/local/local_worker_manager.py
```diff
@@ -0,0 +1,230 @@
...
+      # start the worker and the message pump
+      self._process.start()
+      # the observer must follow the process start, otherwise join() raises.
+      self._observer.start()
+      self._pump.start()
+
+    def _msg_pump(self):
+      while True:
+        task_result = self._receive.get()
+        if not task_result:
```
yundiqian, line 148: task_result is None

compiler_opt/distributed/local/local_worker_manager.py
```diff
@@ -0,0 +1,230 @@
...
+    def __getattr__(self, name) -> Callable[[Any], concurrent.futures.Future]:
+      result_future = concurrent.futures.Future()
+
+      with self._msgidlock:
+        msgid = self._msgid
+        self._msgid += 1
+
+      def remote_call(*args, **kwargs):
+        with self._lock:
+          if self._is_stopped():
```
yundiqian, line 177: how about the time gap between the process is shut down but the None info is not received yet? probably just check whether the process is still alive

mtrofin: This relies on one signal instead of 2, so less triggers for transitioning the states of the stub. There's not much of an inefficiency happening: if the process was killed and these lines execute first, we just enqueue the message and then the message pump will cancel it. It's all happening at the end of training, so any inefficiency - albeit in this case probably not even measurable - doesn't matter much.

compiler_opt/distributed/local/local_worker_manager.py
mtrofin: process.kill may raise for various reasons, like the process already died, etc. all we want to do here is say kill -9, no questions asked, all we care is the net result is the process is killed.

compiler_opt/distributed/local/local_worker_manager.py
mtrofin: done

compiler_opt/distributed/local/local_worker_manager.py
mtrofin: done

PR #34, opened 2022-06-29
Add caching to input pipeline

Northbadge:
Caching the results pre-shuffling prevents TF from recomputing the pre-shuffle batches from scratch repeatedly. 
A speedup of about >5x is expected for the input pipeline, and similarly for the training cycle times as well.

### DIFF
--- a/compiler_opt/rl/train_locally.py
+++ b/compiler_opt/rl/train_locally.py
@@ -126,7 +126,7 @@ def train_eval(agent_name=constant.AgentName.PPO,
       train_sequence_length=train_sequence_length)
 
   def sequence_example_iterator_fn(seq_ex: List[str]):
-    return iter(dataset_fn(seq_ex).repeat())
+    return iter(dataset_fn(seq_ex).repeat().prefetch(tf.data.AUTOTUNE))
 
   reward_stat_map = collections.defaultdict(lambda: None)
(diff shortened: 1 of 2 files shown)


compiler_opt/rl/train_locally.py
```diff
@@ -126,7 +126,7 @@ def train_eval(agent_name=constant.AgentName.PPO,
       train_sequence_length=train_sequence_length)
 
   def sequence_example_iterator_fn(seq_ex: List[str]):
-    return iter(dataset_fn(seq_ex).repeat())
+    return iter(dataset_fn(seq_ex).repeat().prefetch(tf.data.AUTOTUNE))
```
yundiqian, line 129: does where the prefetch is affect the performance?

yundiqian (approved):
is the performance now similar or slow to the ABCCBA solution?

compiler_opt/rl/train_locally.py
Northbadge: > is the performance now similar or slow to the ABCCBA solution?

about 2~3x slower, but now we're talking <1s (ABCCBA) vs 2~3s for this

yundiqian: ic, yeah, i guess it makes a lot of sense to have the prefetch at the end of the dataset pipeline

PR #45, opened 2022-07-06
Refactor file_paths into ModuleSpec

Northbadge:
The file_paths tuple is replaced with a ModuleSpec dataclass.
This has the benefit of not recomputing the entire command line options on each compilation, and also simplifies/unifies runners' inputs.

- Adds ThinLTO support for the inlining runner.

- Corpus-specific command options overrides can now be specified. (for linker invoked ThinLTO corpus training)

- Extension to support corpus-specific command option additions should also be (trivially) implementable in the future. (b/233935329)

- Fix generate_default_trace regression from 09b41a8 (missing delete_flags)


compiler_opt/rl/problem_configuration.py
```diff
@@ -100,6 +105,143 @@ def get_nonnormalized_features(self) -> Iterable[str]:
...
+      additional_flags: tuple of clang flags to add.
+      delete_flags: tuple of clang flags to remove.
+    """
+    raise NotImplementedError
+
+  @staticmethod
+  def _get_module_specs(data_path: str, additional_flags, delete_flags,
+                        xopts) -> List[ModuleSpec]:
+    module_paths: List[str] = get_module_paths(data_path)
+
```
yundiqian, line 128: everything below here should probably in another .py file, probably together with the current adt.py (renamed to module_spec.py)

and also need to add unit test for these functions

compiler_opt/rl/problem_configuration.py
```diff
@@ -100,6 +105,143 @@ def get_nonnormalized_features(self) -> Iterable[str]:
...
+    xopts.update({'output': ['-o', '{path:s}']})
+
+    module_specs: List[ModuleSpec] = []
 
-def is_thinlto(module_paths: Iterable[str]) -> bool:
+    # Future: cmd_override could have per-module overrides if needed
+    cmd_override = get_cmd_override(meta, critical=not has_cmd)
+
+    for module_path in module_paths:
```
yundiqian, line 143: how long does this for loop take? if it's taking too long, may need multithreading/multiprocessing here

compiler_opt/rl/compilation_runner_test.py
```diff
@@ -28,6 +28,8 @@
 
 from compiler_opt.rl import compilation_runner
 from compiler_opt.rl import constant
+from compiler_opt.rl.adt import ModuleSpec
+from compiler_opt.rl.problem_configuration import get_command_line_for_bundle
```
yundiqian, line 32: the test for get_command_line_for_bundle related should be in the *_test.py file that * == the module the function belongs to

yundiqian (commented):
first few comments, still looking

compiler_opt/rl/problem_configuration.py
yundiqian: _xopts looks trivial, but anyway if it's 7s, then not worth doing it.

compiler_opt/rl/problem_configuration.py
```diff
@@ -100,6 +105,143 @@ def get_nonnormalized_features(self) -> Iterable[str]:
...
+        os.path.join(data_path, 'metadata.json'), 'r', encoding='utf-8') as f:
+      return json.load(f)
+  except FileNotFoundError:
+    logging.info('metadata.json couldn\'t be found.')
+    return {}
+
+
+def get_cmd_override(meta: Dict[any, any],
+                     critical=False) -> Union[List[str], None]:
+  if 'global_command_override' in meta:
```
yundiqian, line 182: what's the use case for global_command_override?

compiler_opt/rl/problem_configuration.py
```diff
@@ -100,6 +105,143 @@ def get_nonnormalized_features(self) -> Iterable[str]:
...
+    logging.info('global_command_override will be used instead of .cmd files')
+    return meta['global_command_override']
+  elif critical:
+    logging.fatal('global_command_override couldn\'t be found but is needed')
+    sys.exit(1)
```
yundiqian, line 187: logging.fatal will already exit

compiler_opt/rl/problem_configuration.py
```diff
@@ -100,6 +105,143 @@ def get_nonnormalized_features(self) -> Iterable[str]:
...
+    # {additional|delete}_flags here
+    meta = load_metadata(data_path)
+
+    xopts.update({'output': ['-o', '{path:s}']})
+
+    module_specs: List[ModuleSpec] = []
 
-def is_thinlto(module_paths: Iterable[str]) -> bool:
+    # Future: cmd_override could have per-module overrides if needed
+    cmd_override = get_cmd_override(meta, critical=not has_cmd)
```
yundiqian, line 141: a way to make the logic clearer:

you can just load the cmd_override here and set it to None if not exists in metadata, then delegate the logic of processing them in get_command_line_for_bundle provided the .cmd file and cmd_override, then the if/else/warning raise/error raise would be much clearer

compiler_opt/rl/adt.py
```diff
@@ -0,0 +1,44 @@
...
+import logging
+from typing import List
+import sys
+
+
+@dataclass(frozen=True)
+class ModuleSpec:
+  """Dataclass describing an input module and its compilation command options.
+  """
+  _exec_cmd: List[str]
```
yundiqian, line 27: can be tuple

compiler_opt/rl/adt.py
```diff
@@ -0,0 +1,44 @@
...
+  _xopts: dict[str, List[str]]
```
yundiqian, line 28: Tuple as well

compiler_opt/rl/problem_configuration.py
Northbadge: it'll be for training on corpora extracted from the linker (in non-distributed ThinLTO cases); since we're having clang invoke ThinLTO in the training cycles instead, we can't reuse the original command as-is. Currently I'm still missing `ir_extractor.py` changes to make the workflow easier, so I haven't wrote instructions for release on GitHub yet. 

There's instructions in my internal design doc to get it running for now though

compiler_opt/rl/inlining/__init__.py
```diff
@@ -14,15 +14,17 @@
 # limitations under the License.
 """Implementation of the 'inlining for size' problem."""
 
+from typing import Tuple, List
 import gin
 
-from compiler_opt.rl import problem_configuration
+from compiler_opt.rl.problem_configuration import ProblemConfiguration
```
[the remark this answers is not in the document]
Northbadge: I've switched these back to importing the modules, but for `module_spec` it feels redundant to have `module_spec.ModuleSpec` so I turned it into `adt.ModuleSpec` (using the __init__) file. Resolved all the other comments as well, thanks for the review!

compiler_opt/rl/compilation_runner.py
```diff
@@ -331,15 +285,15 @@ def signaler():
 
   def collect_data(
       self,
-      file_paths: Tuple[str, ...],
+      module_spec: adt.ModuleSpec,
```
mtrofin, line 288: `adt` is too generic - abstract data types? How about a `corpus` module; then we can add a corpus reader encapsulate a source of `ModuleSpec`s. That will, among other things, simplify how we know it is or not thinlto, because that is a corpus-wide property.

compiler_opt/rl/compilation_runner_test.py
```diff
@@ -107,7 +108,7 @@ def test_policy(self, mock_compile_fn):
     runner = compilation_runner.CompilationRunner(
         moving_average_decay_rate=_MOVING_AVERAGE_DECAY_RATE)
     data = runner.collect_data(
-        file_paths=('bc', 'cmd'),
+        module_spec=adt.ModuleSpec(('-O2',), {}, 'dummy'),
```
mtrofin, line 111: can you kwargs, more readable

compiler_opt/rl/inlining/__init__.py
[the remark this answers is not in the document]
Northbadge: Switch to `corpus.py` so this is resolved now

compiler_opt/rl/compilation_runner_test.py
```diff
@@ -107,7 +108,8 @@ def test_policy(self, mock_compile_fn):
     runner = compilation_runner.CompilationRunner(
         moving_average_decay_rate=_MOVING_AVERAGE_DECAY_RATE)
     data = runner.collect_data(
-        file_paths=('bc', 'cmd'),
+        module_spec=corpus.ModuleSpec(
+            _exec_cmd=('-O2',), _xopts={}, name='dummy'),
```
mtrofin, line 112: rename `_exec_cmd` to `exec_cmd`, because it's now part of the public interface; `_xopts` to `extra_opts` (more verbose, but more clear, too)

compiler_opt/rl/train_locally.py
```diff
@@ -100,23 +99,14 @@ def train_eval(agent_name=constant.AgentName.PPO,
...
-        os.path.join(FLAGS.data_path, name.rstrip('\n')) for name in f
-    ]
-
-    if not problem_configuration.is_thinlto(module_paths):
-      file_paths = [(path + '.bc', path + '.cmd') for path in module_paths]
-    else:
-      file_paths = [(path + '.bc', path + '.cmd', path + '.thinlto.bc')
-                    for path in module_paths]
+  logging.info('Loading module specs from corpus')
+  module_specs = problem_config.get_spec_type().get(FLAGS.data_path,
```
mtrofin, line 103: I think you can further decouple the spec thing from the corpus and problem solvers and so on: just let it be a problem-internal consideration.

compiler_opt/rl/corpus.py
```diff
@@ -0,0 +1,180 @@
...
+def _has_cmd(module_paths: Iterable[str]) -> bool:
+  return tf.io.gfile.exists(next(iter(module_paths)) + '.cmd')
+
+
+def _load_module_paths(data_path, module_paths_path: str) -> List[str]:
+  with open(module_paths_path, 'r', encoding='utf-8') as f:
+    ret = [os.path.join(data_path, name.rstrip('\n')) for name in f]
+    if len(ret) == 0:
+      logging.fatal('%s is empty.', module_paths_path)
+      raise ValueError
```
yundiqian, line 114: logging.fatal kills everything so don't need to raise an error here

compiler_opt/rl/corpus.py
```diff
@@ -0,0 +1,180 @@
...
+            'Command line addition of \'%s\' requested '
+            'but no such extra option exists.', k)
+        raise ValueError
+    return ret_cmd
+
+  @classmethod
+  def _get(cls, data_path: str, additional_flags: Tuple[str, ...],
+           delete_flags: Tuple[str, ...], xopts: Dict[str, Tuple[str, ...]]):
+    module_paths: List[str] = _load_module_paths(
+        data_path, os.path.join(data_path, 'module_paths'))
```
yundiqian, line 55: only pass data_path to _load_module_paths function

compiler_opt/rl/corpus.py
```diff
@@ -0,0 +1,180 @@
...
+
+    # This takes ~7s for 30k modules
+    for module_path in module_paths:
+      exec_cmd = _load_and_parse_command(
+          cmd_file=(module_path + '.cmd') if has_cmd else None,
+          ir_file=module_path + '.bc',
+          thinlto_file=(module_path + '.thinlto.bc') if is_thinlto else None,
+          additional_flags=additional_flags,
+          delete_flags=delete_flags,
+          cmd_override=cmd_override)
```
yundiqian, line 82: easier to just pass thw whole metadata to _load_and_parse_command so further information can be directly extracted from meta if needed in the future

compiler_opt/rl/inlining/__init__.py
```diff
@@ -14,11 +14,13 @@
 # limitations under the License.
 """Implementation of the 'inlining for size' problem."""
 
+from typing import Tuple, List
 import gin
 
 from compiler_opt.rl import problem_configuration
 from compiler_opt.rl.inlining import config
 from compiler_opt.rl.inlining import inlining_runner
+from compiler_opt.rl.inlining import inlining_spec
```
yundiqian, line 23: spec->module_spec everywhere cause spec is too general so it can cause confusion

compiler_opt/rl/corpus.py
yundiqian: so either doing logging.fatal or raise ValueError with proper error message, recommend to raise ValueErrors to be consistent with the codebase

compiler_opt/rl/corpus.py
```diff
@@ -0,0 +1,180 @@
...
+    cmdline = list(cmd_override)
+  elif cmd_file is not None:
+    with open(cmd_file, encoding='utf-8') as f:
+      option_iterator = iter(f.read().split('\0'))
+      option = next(option_iterator, None)
+  else:
+    logging.fatal('.cmd file not available and no command override specified.')
+    raise ValueError
+
+  while option:
```
yundiqian, line 160: move the code from option->cmdline above in elif loop to make the logic clearer --- cmdline is prepared after the if/else loop, either coming from cmd_override or reading from cmd file

compiler_opt/rl/corpus.py
```diff
@@ -0,0 +1,180 @@
...
+    option = next(option_iterator, None)
+
+  cmdline.extend(['-x', 'ir', ir_file])
+
+  if thinlto_file:
+    cmdline.append('-fthinlto-index=' + thinlto_file)
+    cmdline += ['-mllvm', '-thinlto-assume-merged']
+
+  if cmd_override is None:
+    cmdline.extend(additional_flags)
```
yundiqian, line 175: ('-mllvm', '-enable-ml-inliner=development') is added to the additional_flags, it is not needed in the cmd_override case?

compiler_opt/rl/corpus.py
```diff
@@ -0,0 +1,180 @@
...
+    # Ensure that -cc1 is always present
+    if cmdline[0] != '-cc1':
```
yundiqian, line 177: is it supposed to happen?

compiler_opt/rl/train_locally.py
yundiqian: currently it seems to be a problem-internal consideration and decoupled well, how the further decoupling looks like?

compiler_opt/rl/corpus.py
Northbadge: ah ok, I put both since logging.fatal didn't seem kill the tests

compiler_opt/rl/corpus.py
Northbadge: Only considering non-linker extracted corpora, i.e. when there's `.cmd` files, the command line inside the `.cmd` file should be run with `-cc1`. Will switch it to `if cmd_file is not None` though, safer that way

compiler_opt/rl/corpus.py
Northbadge: good catch, with the new commit I moved the -enable-ml-inliner flag to somewhere else so it'll be added no matter what now

compiler_opt/rl/train_locally.py
Northbadge: they're back in `*_runner` now

mtrofin:
@Northbadge, capturing offline chat - are you planning to have a smaller initial PR where the original `Tuple[str,...]` becomes `ModuleSpec` and nothing else?

Northbadge:
> @Northbadge, capturing offline chat - are you planning to have a smaller initial PR where the original `Tuple[str,...]` becomes `ModuleSpec` and nothing else?

Don't think so; I thought that was more geared towards introducing the new thing in one PR and switching over to the new thing in another PR

mtrofin:
> 

It may help doing the simple Tuple->ModuleSpec replacement first (and rebasing this, and maybe splitting this in a few stages) because it would ease understand the diffs. The first Tuple->ModuleSpec would basically be straight-forward, and not much to analyze, and also solve 90% of the problem (basically, much better readability)

PR #47, opened 2022-07-10
Benchmarking utilities

boomanaiden154:
After talking with Mircea over slack, the decision was made to look at upstreaming my benchmarking utilities (for chromium and the llvm test suite). I have separated everything over a couple of different python files. There are three scripts that get added in `benchmark_chromium.py`, `benchmark_llvm_test_suite.py`, and `list_gtests.py`.
* `benchmark_chromium.py` contains all of the facilities to run the chromium gtest executables, also automating model integration into LLVM and compiling the tests with different models.
* `benchmark_llvm_test_suite.py` is quite similar to `benchmark_chromium.py`, but focused on the LLVM test suite. By default, it executes most of the MicroBenchmarks from the llvm test suite repository, and also automates the building of the test suite using a two-stage PGO build, as well as compiling models into LLVM.
* `list_gtests.py` is just a utility script that was used for collecting all of the test names in the chromium gtest executables. It doesn't need to be used unless you're preparing your own chromium test suite as I have prepared a default corpus of tests (methodology for that available [here](https://github.com/boomanaiden154/gsoc2022-notes/blob/master/july/3/analysis1.ipynb)), but that might need to be done in the future and I had already written the utility, so it should probably be included.
Then there are the two utility files:
* `benchmarking_utils.py` - contains common benchmarking utils such as the function to build LLVM, and also the microbenchmark runner
* `gtest_executable_utils.py` - contains a bunch of utilities to work with/run gtest executables.

The instrumentation for the microbenchmarks is built into google benchmark, so that is used for the llvm test suite. Since chromium only uses google test and there is no built in perf counter instrumentation, I'm just using perf stat to grab performance counters. It's not perfect, but based on my analysis the run to run variability is quite low (definitely good enough to see the changes that we're expecting) and seem to be fairly valid as well (showing a reduction in loads over the default heuristic). 

Design decisions to note:
* I'm compiling the model into LLVM rather than just using the development mode advisor as we don't need any of the features that the development mode advisor has, and since in release mode the model is ahead of time compiled, it should also be faster (which is important as the chromium benchmark compile takes several hours on a decent sized machine and has about 60000 build steps).
* I haven't written any unit tests for these files, as a major component of their functionality is how they interact with the environment around them. I'm planning on writing one or two integration tests that can be used to verify functionality, although we won't be able to easily plug those into CI.
* I'm currently using `joblib` in order to run some of the tests in parallel. Based on some of my testing some of the perf counters are reliable enough even when run in parallel (data/analysis availabe [here](https://github.com/boomanaiden154/gsoc2022-notes/blob/master/july/4/analysis.ipynb)). However, `joblib` isn't used anywhere else in the project as the built-in `multiprocessing` library is. I can refactor this if needed. The chromium tests by default take about 3-4 hours to run single threaded, so running them in parallel provides an immense speed-up.

TODO:
* [ ] Add type annotations to all of the new functions/classes
* [ ] Add in documentation on running the new benchmarking scripts
* [ ] Write integration test shell scripts
* [ ] Adjust output of scripts to match what the existing benchmark compare function is expecting


boomanaiden154:
Also, the `check_license.sh` script seems to be checking for the header with the copyright year set to 2020. Should I just use that header in my files or should I set the copyright year to 2022 since that is when they were written?

docs/benchmarking.md
```diff
@@ -0,0 +1,183 @@
...
+this with caution as it can add a lot of noise to your benchmarks depending
+upon what specifically you are doing.
+* `--perf_counters` - similar to the llvm test suite perf counters, but instead
+of being in the libpfm format, they're perf counters as listed in `perf list`.
+* `--test_description` - Can be declared multiple times if you have custom test
+descriptions that you want to run, but the default works well, covers a broad
+portion of the codebase, and has been specifically designed to minimize run
+to run variability.
+
+## Generating Chromium Test Descriptions
```
yundiqian, line 147: make it ### so that it is under Benchmarking Chromium, and also add the explanation about it in your PR description to the md file here

boomanaiden154: Adjusted the heading level and added in all the relevant info from the PR description into the file.

compiler_opt/benchmark/chromium_test_descriptions/base_perftests.json
```diff
@@ -0,0 +1,163 @@
+{
```
mtrofin, line 1: how do we evolve this file over time?

mtrofin:
lgtm. Thanks for doing this!

PR #52, opened 2022-07-13
Refactor file_paths into ModuleSpec 

Northbadge:
Rewrote the commit history of #45 and didn't want to wipe that PR's comments out, so making a new PR instead.

First commit crudely switches from file_paths to ModuleSpec (and has regressions), second commit completes the switch with new functions/flow.

- The file_paths tuple is replaced with a ModuleSpec dataclass.
This has the benefit of not recomputing the entire command line options on each compilation, and also simplifies/unifies runners' inputs.

- Adds ThinLTO support for the inlining runner.

- Corpus-specific command options overrides can now be specified via metadata.json. (for linker invoked ThinLTO corpus training)

- Extension to support corpus-specific command option additions should also be (trivially) implementable in the future. (b/233935329)

- Fix generate_default_trace regression from https://github.com/google/ml-compiler-opt/commit/09b41a8bda02ebfe47c5e9bc8149dbf9537b9fe4 (missing delete_flags)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 __pycache__
 .vscode
+.idea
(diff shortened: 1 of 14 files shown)


Northbadge:
splitting into even more incremental changes, so moving to the #57 and onwards PR chain

PR #53, opened 2022-07-13
Refactored train_eval to use num_modules from gin configs

boomanaiden154:
This patch refactors the passed `num_modules` flag in `train_locally.py` into a setting in the gin configs while preserving the default value of 100. This is a parameter that can have a major impact on the performance of a trained model (at least with the regalloc case) and thus should probably be specified in the gin configs (when it deviates from the default value of 100) for the sake of reproducibility. 

Eg for the regalloc case, the result with `num_modules` set to 512 is a graph of the mean reward per iteration which displays the characteristic shape of a model that is training properly (only trained for 100k iterations):
image
 And this is the same model trained for 600k iterations with `num_modules` set to 100:
image

### DIFF
--- a/compiler_opt/rl/regalloc/gin_configs/ppo_nn_agent.gin
+++ b/compiler_opt/rl/regalloc/gin_configs/ppo_nn_agent.gin
@@ -13,6 +13,7 @@ include 'compiler_opt/rl/regalloc/gin_configs/network.gin'
 train_eval.agent_name=%constant.AgentName.PPO
 train_eval.warmstart_policy_dir=''
 train_eval.num_policy_iterations=3000
+train_eval.num_modules=512
 train_eval.num_iterations=200
 train_eval.batch_size=256
 train_eval.train_sequence_length=16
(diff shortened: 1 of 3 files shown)


boomanaiden154:
@yundiqian Do we need to adjust the `num_modules` setting in any of the inlining gin configs? Also, does the value of 512 look good? On Monday I believe you said that you used the value 488, but the difference is small and I'd probably prefer a power of 2/10 for aesthetic reasons since it probably doesn't matter that much.

compiler_opt/rl/regalloc/gin_configs/ppo_nn_agent.gin
```diff
@@ -19,6 +19,7 @@ train_eval.train_sequence_length=16
 train_eval.deploy_policy_name='saved_collect_policy'
 train_eval.moving_average_decay_rate=0.8
 train_eval.use_random_network_distillation=False
+train_eval.num_modules=512
```
yundiqian, line 22: also add the flag to inlining ppo gin file

yundiqian:
> 512 looks good



> @yundiqian Do we need to adjust the `num_modules` setting in any of the inlining gin configs? Also, does the value of 512 look good? On Monday I believe you said that you used the value 488, but the difference is small and I'd probably prefer a power of 2/10 for aesthetic reasons since it probably doesn't matter that much.

compiler_opt/rl/regalloc/gin_configs/ppo_nn_agent.gin
boomanaiden154: With the same value?

PR #61, opened 2022-07-15
Make generate_default_trace fail-fast when hitting a bug

mtrofin:
If the worker process hits an unrecoverable error, we stop processing
instead of blocking indefinitely.

### DIFF
--- a/compiler_opt/tools/generate_default_trace.py
+++ b/compiler_opt/tools/generate_default_trace.py
@@ -21,7 +21,7 @@ import queue
 import random
 import re
 import subprocess
-from typing import Dict, List, Optional, Tuple, Type  # pylint:disable=unused-import
+from typing import Dict, List, Optional, Union, Tuple  # pylint:disable=unused-import
 
 from absl import app
 from absl import flags
@@ -66,8 +66,10 @@ _GIN_BINDINGS = flags.DEFINE_multi_string(
     'gin_bindings', [],
     'Gin bindings to override the values set in the config files.')
 
-ResultsQueueEntry = Optional[Tuple[str, List[str],
-                                   Dict[str, compilation_runner.RewardStat]]]
+ResultsQueueEntry = Union[Optional[Tuple[str, List[str],
+                                         Dict[str,
+                                              compilation_runner.RewardStat]]],
+                          BaseException]
 
 
 def get_runner() -> compilation_runner.CompilationRunner:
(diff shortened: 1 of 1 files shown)


compiler_opt/tools/generate_default_trace.py
```diff
@@ -206,6 +214,8 @@ def main(_):
                                     total_failed_examples, total_work)
 
         results = results_queue.get()
+        if isinstance(results, BaseException):
+          logging.fatal(results)
```
Northbadge, line 218: `logging.fatal` doesn't insta-exit in the test environment, it errors out a couple lines down when trying to unpack the exception instead. Could switch to `logging.error` and reraise the exception

yundiqian: probably just raise the error

yundiqian (approved):

compiler_opt/tools/generate_default_trace.py
```diff
@@ -97,35 +99,41 @@ def worker(policy_path: str, work_queue: 'queue.Queue[corpus.ModuleSpec]',
...
+          if not m.match(k):
+            continue
+          new_reward_stats[k] = data.reward_stats[k]
+          new_sequence_examples.append(sequence_example)
+        results_queue.put(
+            (module_spec.name, new_sequence_examples, new_reward_stats))
+      except (subprocess.CalledProcessError, subprocess.TimeoutExpired,
+              RuntimeError):
+        logging.error('Failed to compile %s.', module_spec.name)
+        results_queue.put(None)
```
yundiqian, line 134: why not have the except her?

yundiqian: i mean do not really need another try except outside, just have try except except here to queue the error to result queue

compiler_opt/tools/generate_default_trace.py
mtrofin: sgtm. left a comment to make sure we don't start catching that

compiler_opt/tools/generate_default_trace.py
yundiqian: hmm...actually it is a bit tricky. so if the compilation fails due to some flag issue (the case you'd want to capture and kill early), it is going to be one of the subprocess.CalledProcessError, subprocess.TimeoutExpired,RuntimeError so it won't raise the BaseException.

one alternative just to have the receiver check on how many None it receives. for whatever reason, if it receives "too much" Nones, e.g., 50%, it kills the code

mtrofin: This is not about that. The compilation error cases are treated together and a report is generated. This is about a bug like we change parameter names in compilation_runner.

basically, compilation failures are expected - timeouts, one-off issue in the corpus, whatever. If the user sees what is unacceptable to them as a number of failed compilation tasks, they can investigate.

bugs like I mentioned are not expected. There's no `analog` thing there (i.e. degrees of allowance), it's straight-out failure.

yundiqian: ic, so essentially you are only trying to catch the error in 'runner = get_runner()', it makes sense then.

i'm wondering why the get_runner() is moved to inside the multiprocessing?

PR #62, opened 2022-07-17
Make the tensorboard data interval adjustable

boomanaiden154:
Currently, all of the tensorboard variables get updated every
single iteration. This is quite costly in terms of performance. I'm
seeing about a 25% performance improvement by just running this once
every 100 iterations instead of running it every iteration. The data
showing up in tensorboard will definitely be more sparse, but when
running everything locally, it seems to all appear just fine, and I'm not sure
getting data down to the iteration level is quite necessary, especially
given the large performance penalty.

I implemented this as an optional flag that can be passed from the command
line with what I believe is a reasonable default (once every 100 iterations), 
so it can be easily adjusted by the user if necessary while also being
a completely reasonable default for good tensorboard data/little
performance loss.

### DIFF
--- a/compiler_opt/rl/trainer.py
+++ b/compiler_opt/rl/trainer.py
@@ -50,7 +50,8 @@ class Trainer(object):
       # Params for summaries and logging
       checkpoint_interval=10000,
       log_interval=100,
-      summary_interval=1000,
+      summary_log_interval=100,
+      summary_export_interval=1000,
       summaries_flush_secs=10):
     """Initialize the Trainer object.
 
@@ -62,8 +63,10 @@ class Trainer(object):
       checkpoint_interval: int, the training step interval for saving
         checkpoint.
       log_interval: int, the training step interval for logging.
-      summary_interval: int, the training step interval for exporting to
-        tensorboard.
+      summary_log_interval: the number of steps in between logging metrics
+        to tensorboard.
+      summary_export_interval: int, the training step interval for exporting
+        to tensorboard.
       summaries_flush_secs: int, the seconds for flushing to tensorboard.
     """
     self._root_dir = root_dir
@@ -71,7 +74,8 @@ class Trainer(object):
     self._random_network_distillation = random_network_distillation
     self._checkpoint_interval = checkpoint_interval
     self._log_interval = log_interval
-    self._summary_interval = summary_interval
+    self._summary_log_interval = summary_log_interval
+    self._summary_export_interval = summary_export_interval
 
     self._summary_writer = tf.summary.create_file_writer(
         self._root_dir, flush_millis=summaries_flush_secs * 1000)
@@ -108,6 +112,7 @@ class Trainer(object):
     self._start_time = time.time()
     self._last_checkpoint_step = 0
     self._last_log_step = 0
+    self._summary_last_log_step = 0
 
   def _initialize_metrics(self):
     """Initializes metrics."""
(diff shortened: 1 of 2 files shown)


compiler_opt/rl/trainer.py
```diff
@@ -51,7 +51,8 @@ def __init__(
       checkpoint_interval=10000,
       log_interval=100,
       summary_interval=1000,
-      summaries_flush_secs=10):
+      summaries_flush_secs=10,
+      tensorboard_data_interval=100):
```
yundiqian, line 55: looks a good idea to set it to be 100 here, let's just make it as default

yundiqian: and call it summary_log_interval to be consistent and move it between log_interval and summary_interval

also add it to Args in docstring below

yundiqian: no need to add a flag cause users can set it with gin config if they really want to change it

compiler_opt/rl/trainer.py
```diff
@@ -50,6 +50,7 @@ def __init__(
       # Params for summaries and logging
       checkpoint_interval=10000,
       log_interval=100,
+      summary_log_interval=100,
       summary_interval=1000,
```
yundiqian, line 54: nit: can you change it to summary_export_interval to be more clear?

compiler_opt/rl/trainer.py
```diff
@@ -77,6 +80,9 @@ def __init__(
         self._root_dir, flush_millis=summaries_flush_secs * 1000)
     self._summary_writer.set_as_default()
 
+    self._summary_log_interval = summary_log_interval
```
yundiqian, line 83: nit: move it above between self._log_interval = log_interval and self._summary_interval = summary_interval

compiler_opt/rl/trainer.py
```diff
@@ -77,6 +80,9 @@ def __init__(
...
+    self._summary_log_last_data_collection = 0
```
yundiqian, line 84: nit: move it below after self._last_log_step = 0

compiler_opt/rl/trainer_test.py
```diff
@@ -92,7 +92,9 @@ def test_training(self):
         tf.compat.v1.train.AdamOptimizer(),
         num_outer_dims=2)
     test_trainer = trainer.Trainer(
-        root_dir=self.get_temp_dir(), agent=test_agent)
+        root_dir=self.get_temp_dir(),
+        agent=test_agent,
+        tensorboard_data_interval=1)
```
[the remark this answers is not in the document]
boomanaiden154: It will fail on this revision because I forgot to change the variable name here (will fix). I forget the exact details, but with all the mocking that is setup for that test, I think it only runs ten iterations or something and then expects tensorboard data to be written after that, which it isn't getting because the new default summary export interval is significantly longer. I can rework the test if you want, but I felt that just changing the parameter was sufficient while still having the test touch on all the functionality that it needs to.

compiler_opt/rl/trainer.py
yundiqian: sorry for the confusion...github is not clear in terms of what line i choose -_-!

i mean to have summary_log_interval for the ones you added, and rename summary_interval to be summary_export_interval to make these two flags more clear

compiler_opt/rl/trainer_test.py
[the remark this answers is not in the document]
yundiqian: yeah, just changing the parameter is good enough, the mock thing in the unit test is intended to make it a "unit test" :)

compiler_opt/rl/trainer.py
yundiqian: not needed cause we don't really expect to change these values

compiler_opt/rl/trainer.py
boomanaiden154: Should be fixed now. I should probably also pay more attention to what line a comment indicates (since it's the bottom one), and also think about what I'm doing a little bit more. :p

compiler_opt/rl/trainer.py
```diff
@@ -108,6 +112,7 @@ def __init__(
     self._start_time = time.time()
     self._last_checkpoint_step = 0
     self._last_log_step = 0
+    self._summary_last_log = 0
```
yundiqian, line 115: nit: _summary_last_log_step

yundiqian (approved):

PR #63, opened 2022-07-18
Move corpus related functions to corpus.py

Northbadge:
- ModuleSpec.cmd() is made to be a proxy of get_command_line_for_bundle()


Northbadge:
> Peeking forward a bit, it may be better to first introduce a `Corpus` abstraction and its implementation for directory-based, distributed thinlto or not cases, which can be asked to generate ModuleSpec objects; then shortly thereafter we have the lld-based (i.e. 'local' thinlto) case, with its own specifics, such as the way it'd populate the command line of a `ModuleSpec` is different.
> 
> By doing this, it can help tease out design decisions. For example: with the picture above in mind, it's the corpus, not the ModuleSpec, that populates its command line (and thinlto- related artifacts). But a consumer would simply care about: 1) what's the input file or files, 2) what's the command line.
> 
> There's also, down the road, the fact that for afdo or csfdo cases, we want to make sure that 1) the corpus is aware of the shared file (the profile).
> 
> So probably the notions of corpus and module spec kind of go hand in hand, hence my suggestion to also add some corpus notion, and then the picture would shape up more clearly. - wdyt?

Current plan is this -> corpus.read() instead of reading the .cmd every time -> introduce metadata.json which modifies the output of corpus.read() (along with lld-thinlto corpus support) -> deprecate module_paths (simultaneous with extract_ir) -> introduce Corpus object (with more functions) and simultaneously etc. ModuleSpec changes.

Implementing the corpus abstraction first seems like skipping a several steps forwards, which I thought we were trying to avoid

compiler_opt/rl/corpus.py
```diff
@@ -14,12 +14,106 @@
...
+  _exec_cmd: Tuple[str, ...] = ()
+
+  def cmd(self) -> List[str]:
+    """Retrieves the compiler execution options.
+    """
+    # Using a getter as this will be modified in the future
+    return list(self._exec_cmd)
+
+
+def read(
```
yundiqian, line 39: something like build_modulespecs_from_datapath, being more descriptive of what the function do.

compiler_opt/rl/corpus.py
```diff
@@ -14,12 +14,106 @@
...
+def _has_thinlto_index(module_paths: Iterable[str]) -> bool:
+  return tf.io.gfile.exists(next(iter(module_paths)) + '.thinlto.bc')
+
+
+def _load_module_paths(data_path) -> List[str]:
+  module_paths_path = os.path.join(data_path, 'module_paths')
+  with open(module_paths_path, 'r', encoding='utf-8') as f:
+    ret = [os.path.join(data_path, name.rstrip('\n')) for name in f]
+    if len(ret) == 0:
+      logging.error('%s is empty.', module_paths_path)
```
yundiqian, line 72: raise ValueError("...") and no need for logging.error if this error is intended to break the code execution

compiler_opt/rl/corpus.py
```diff
@@ -14,12 +14,106 @@
...
+      raise ValueError
+    return ret
+
+
+def _load_and_parse_command(
+    cmd_file: str,
```
yundiqian, line 78: nit: revert ir_file and cmd_file to be consistent of elsewhere in the codebase

compiler_opt/rl/corpus.py
```diff
@@ -14,12 +14,106 @@
...
+
+  Returns:
+    The argument list to pass to the compiler process.
+  """
+  cmdline = []
+
+  with open(cmd_file, encoding='utf-8') as f:
+    option_iterator = iter(f.read().split('\0'))
+    option = next(option_iterator, None)
+    while option:
```
yundiqian, line 104: nit while option is not None:

compiler_opt/rl/train_locally.py
```diff
@@ -101,17 +100,10 @@ def train_eval(agent_name=constant.AgentName.PPO,
...
-      encoding='utf-8') as f:
-    lines = f.readlines()
-    has_thinlto = problem_configuration.is_thinlto(
-        [os.path.join(FLAGS.data_path, lines[0].rstrip('\n'))])
-    module_specs = [
-        corpus.ModuleSpec(
-            name=os.path.join(FLAGS.data_path, name.rstrip('\n')),
-            has_thinlto=has_thinlto) for name in lines
-    ]
+  logging.info('Loading module specs from corpus')
```
yundiqian, line 103: nit: '...from corpus.', same below

yundiqian (approved):

compiler_opt/rl/corpus.py
```diff
@@ -14,12 +14,105 @@
...
 
 @dataclass(frozen=True)
 class ModuleSpec:
   """Dataclass describing an input module and its compilation command options.
   """
   name: str
-  has_thinlto: bool = False
+  _exec_cmd: Tuple[str, ...] = ()
+
+  def cmd(self) -> List[str]:
```
mtrofin, line 32: why not just leave the exec_cmd publicly accessible - the object is frozen anyway?

compiler_opt/rl/corpus.py
```diff
@@ -14,12 +14,105 @@
...
+    additional_flags: Tuple[str, ...] = (),
+    delete_flags: Tuple[str, ...] = ()
+) -> List[ModuleSpec]:
+  module_paths: List[str] = _load_module_paths(data_path)
+
+  has_thinlto: bool = _has_thinlto_index(module_paths)
+
+  module_specs: List[ModuleSpec] = []
+
+  # This takes ~7s for 30k modules
```
mtrofin, line 50: could be parallelized?

Northbadge: (multi-threading/processing) overhead meant it only became ~4s with diminishing returns with more workers, didn't seem worth it

compiler_opt/rl/corpus.py
Northbadge: To make sure that everything outside uses `.cmd()`,  so later on more stuff can be added on via `.cmd()` without having to change every reference of `.exec_cmd` -> `.cmd()`

mtrofin: but do we want to add more stuff to cmd? I'm assuming that's about stuff like substitutions. I'd argue those belong outside ModuleSpec - as a stand-alone command line converter or something like that. Basically, to compile, you need an array that's appropriately contextualized to the local location the input(s) and profile file(s) are, model under training, extra domain-specific flags... so a (ModuleSpec, extra options) -> array converter. For separation of concerns reasons, the ModuleSpec can stay agnostic of all that, as a logical holder of the related information that can generate that command line. 

My suggestion is to leave ModuleSpec "dumb", basically - so then you can have exec_cmd as a ctor - initialized frozen property.

PR #65, opened 2022-07-22
Replace module_paths with corpus_description.json

Northbadge:
- This adds 'global_command_override' functionality for lld-thinlto corpora, exposed in metadata.json
- Setting the above in the corpus_description.json file will default the runners to using that as the base command, rather than looking for a .cmd file

### DIFF
--- a/compiler_opt/rl/corpus.py
+++ b/compiler_opt/rl/corpus.py
@@ -14,11 +14,12 @@
 # limitations under the License.
 """ModuleSpec definition and utility command line parsing functions."""
 
+from absl import logging
 from dataclasses import dataclass
-from typing import Iterable, Optional, Tuple, List
+from typing import List, Dict, Tuple
 
+import json
 import os
-import tensorflow as tf
 
 
 @dataclass(frozen=True)
@@ -34,44 +35,54 @@ def build_modulespecs_from_datapath(
     additional_flags: Tuple[str, ...] = (),
     delete_flags: Tuple[str, ...] = ()
 ) -> List[ModuleSpec]:
-  module_paths: List[str] = _load_module_paths(data_path)
+  # TODO: (b/233935329) Per-corpus *fdo profile paths can be read into
+  # {additional|delete}_flags here
+  metadata: Dict[str, any] = _load_metadata(
+      os.path.join(data_path, 'metadata.json'))
+  module_paths = metadata['modules']
+  if len(module_paths) == 0:
+    raise ValueError(f'{data_path}\'s metadata contains no modules.')
 
-  has_thinlto: bool = _has_thinlto_index(module_paths)
+  has_thinlto: bool = metadata['has_thinlto']
+
+  # Future: cmd_override could have per-module overrides if needed
+  if 'global_command_override' in metadata:
+    if metadata['global_command_override'] == ['Please', 'fill', 'options']:
+      raise ValueError('global_command_override in metadata.json not filled.')
+    cmd_override = tuple(metadata['global_command_override'])
+    if len(additional_flags) > 0:
+      logging.warning('Additional flags are specified together with override.')
+    if len(delete_flags) > 0:
+      logging.warning('Delete flags are specified together with override.')
+  else:
+    cmd_override = ()
 
   module_specs: List[ModuleSpec] = []
 
   # This takes ~7s for 30k modules
   for module_path in module_paths:
     exec_cmd = _load_and_parse_command(
-        ir_file=module_path + '.bc',
-        cmd_file=(module_path + '.cmd'),
-        thinlto_file=(module_path + '.thinlto.bc') if has_thinlto else None,
+        module_path=os.path.join(data_path, module_path),
+        has_thinlto=has_thinlto,
         additional_flags=additional_flags,
-        delete_flags=delete_flags)
+        delete_flags=delete_flags,
+        cmd_override=cmd_override)
     module_specs.append(ModuleSpec(name=module_path, exec_cmd=tuple(exec_cmd)))
 
   return module_specs
 
 
-def _has_thinlto_index(module_paths: Iterable[str]) -> bool:
-  return tf.io.gfile.exists(next(iter(module_paths)) + '.thinlto.bc')
-
-
-def _load_module_paths(data_path) -> List[str]:
-  module_paths_path = os.path.join(data_path, 'module_paths')
-  with open(module_paths_path, 'r', encoding='utf-8') as f:
-    ret = [os.path.join(data_path, name.rstrip('\n')) for name in f]
-    if len(ret) == 0:
-      raise ValueError(f'{module_paths_path} is empty.')
-    return ret
+def _load_metadata(metadata_path: str) -> Dict[str, any]:
+  with open(metadata_path, 'r', encoding='utf-8') as f:
+    return json.load(f)
 
 
 def _load_and_parse_command(
-    ir_file: str,
-    cmd_file: str,
-    thinlto_file: Optional[str] = None,
+    module_path: str,
+    has_thinlto: bool,
     additional_flags: Tuple[str, ...] = (),
-    delete_flags: Tuple[str, ...] = ()
+    delete_flags: Tuple[str, ...] = (),
+    cmd_override: Tuple[str, ...] = ()
 ) -> List[str]:
   """Cleans up base command line.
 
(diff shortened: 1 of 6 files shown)


compiler_opt/rl/corpus.py
```diff
@@ -34,77 +35,93 @@ def build_modulespecs_from_datapath(
...
+      os.path.join(data_path, 'metadata.json'))
+  module_paths = metadata['modules']
+  if len(module_paths) == 0:
+    raise ValueError(f'{data_path}\'s metadata contains no modules.')
+
+  has_thinlto: bool = metadata['has_thinlto']
+
+  # Future: cmd_override could have per-module overrides if needed
+  if 'global_command_override' in metadata:
+    if metadata['global_command_override'] == ['Please', 'fill', 'options']:
```
mtrofin, line 50: nit: instead of "please fill options", maybe you can have a token like "UNSPECIFIED" or something like that (more "formal")

compiler_opt/rl/corpus.py
```diff
@@ -34,77 +35,93 @@ def build_modulespecs_from_datapath(
...
+  # {additional|delete}_flags here
+  metadata: Dict[str, any] = _load_metadata(
+      os.path.join(data_path, 'metadata.json'))
+  module_paths = metadata['modules']
+  if len(module_paths) == 0:
+    raise ValueError(f'{data_path}\'s metadata contains no modules.')
+
+  has_thinlto: bool = metadata['has_thinlto']
+
+  # Future: cmd_override could have per-module overrides if needed
```
mtrofin, line 48: We don't have a scenario for that, do we?

compiler_opt/rl/corpus.py
```diff
@@ -34,77 +35,93 @@ def build_modulespecs_from_datapath(
     additional_flags: Tuple[str, ...] = (),
     delete_flags: Tuple[str, ...] = ()
 ) -> List[ModuleSpec]:
-  module_paths: List[str] = _load_module_paths(data_path)
-
-  has_thinlto: bool = _has_thinlto_index(module_paths)
+  # TODO: (b/233935329) Per-corpus *fdo profile paths can be read into
+  # {additional|delete}_flags here
+  metadata: Dict[str, any] = _load_metadata(
+      os.path.join(data_path, 'metadata.json'))
```
mtrofin, line 41: instead of `metadata` for the name, let's use `corpus_description`: it's more specific.

compiler_opt/rl/corpus_test.py
```diff
@@ -25,19 +26,21 @@ class CommandParsingTest(tf.test.TestCase):
 
   def test_thinlto_file(self):
     data = ['-cc1', '-foo', '-bar=baz']
-    argfile = self.create_tempfile(content='\0'.join(data))
+    tempdir = self.create_tempdir()
```
mtrofin, line 29: just curious, why did we need a directory now? just opportunistic improvement? (sgtm, in that case, asking to learn)

compiler_opt/rl/corpus.py
mtrofin: or '<UNSPECFIED>'

compiler_opt/rl/corpus_test.py
Northbadge: ah, more of a I forgot to remove after changing the signature of _load_and_parse_command, will revert

compiler_opt/rl/corpus.py
Northbadge: meant it as a note for "if you ever want per module overrides, here's where to do it", I can only think of if x flag exists, override with y set of options, which sounds plausible enough

mtrofin: I'd remove the comment though - if we hit the scenario in the future, we'll figure it out (i.e. we'll see what exactly the scenario entails, and, surely enough, if it fits with what's here, great).

compiler_opt/rl/corpus.py
Northbadge: were you going to suggest something else?

mtrofin: yes - sorry. <UNSPECIFIED> (not sure what happened)

Northbadge: got it, thanks for the review!

compiler_opt/rl/corpus.py
```diff
@@ -34,77 +35,93 @@ def build_modulespecs_from_datapath(
...
+  corpus_description: Dict[str, Any] = _load_corpus_description(
+      os.path.join(data_path, 'corpus_description.json'))
+  module_paths = corpus_description['modules']
+  if len(module_paths) == 0:
+    raise ValueError(f'{data_path}\'s corpus_description contains no modules.')
+
+  has_thinlto: bool = corpus_description['has_thinlto']
+
+  if 'global_command_override' in corpus_description:
+    if corpus_description['global_command_override'] == ['<UNSPECIFIED>']:
```
mtrofin, line 49: sorry, one more nit: can you define a module constant for the `<UNSPECIFIED>` value, and then use that in extract_ir and elsewhere, rather than the string? it would help maintainability

mtrofin:
lgtm post-last nit, see if @yundiqian has additional feedback and we can then merge - thanks for doing this!

compiler_opt/rl/corpus.py
Northbadge: got it, I put it in `extract_ir.py` though, so that `extract_ir.py` can be run from whatever cwd without trying to import from `compiler_opt`

compiler_opt/rl/corpus.py
```diff
@@ -34,77 +37,94 @@ def build_modulespecs_from_datapath(
     additional_flags: Tuple[str, ...] = (),
     delete_flags: Tuple[str, ...] = ()
 ) -> List[ModuleSpec]:
-  module_paths: List[str] = _load_module_paths(data_path)
-
-  has_thinlto: bool = _has_thinlto_index(module_paths)
+  # TODO: (b/233935329) Per-corpus *fdo profile paths can be read into
+  # {additional|delete}_flags here
+  corpus_description: Dict[str, Any] = _load_corpus_description(
```
yundiqian, line 42: just read the file here? also two lines

compiler_opt/rl/corpus.py
```diff
@@ -34,77 +37,94 @@ def build_modulespecs_from_datapath(
...
+        'global_command_override'] == extract_ir.UNSPECIFIED_OVERRIDE:
+      raise ValueError(
+          'global_command_override in corpus_description.json not filled.')
+    cmd_override = tuple(corpus_description['global_command_override'])
+    if len(additional_flags) > 0:
+      logging.warning('Additional flags are specified together with override.')
+    if len(delete_flags) > 0:
+      logging.warning('Delete flags are specified together with override.')
+  else:
+    cmd_override = ()
```
yundiqian, line 61: nit

cmd_override = ()
if 'global_command_override' in ....

compiler_opt/tools/extract_ir.py
```diff
@@ -333,13 +336,23 @@ def main(argv):
...
-    with open(
-        os.path.join(FLAGS.output_dir, 'module_paths'), 'w',
-        encoding='utf-8') as f:
-      for path in relative_output_paths:
-        if path is not None:
-          f.write(path + '\n')
+  # This comes first rather than later so global_command_override is at the top
+  # of the .json after being written
+  if FLAGS.thinlto_build == 'local':
+    corpus_description = {'global_command_override': UNSPECIFIED_OVERRIDE}
```
yundiqian, line 342: when this will be set the true override commandline? am I missing anything or this is going to be implemented in the next patch?

Northbadge: the user needs to set it in the .json file, I could add an input prompt right here though (which if left blank will default to unspecified)

compiler_opt/rl/corpus.py
Northbadge: slight problem: running `import extract_ir` from train_locally triggers the absl.FLAGS stuff, and we have a name conflict there due to `num_workers` (doing the reverse would be the same result)

compiler_opt/tools/extract_ir.py
yundiqian: do you mean that the user need to manually set it in the json file after running the script?

Northbadge: yes

yundiqian: where does this command line come from? why we cannot have it in the script?

Northbadge: the script is run on a backend-invoked-thinlto corpus (so ThinLTO is invoked by lld, rather than clang). We have a command to re-run lld, but it only works correctly if we want to train on the entire corpus on every iteration. (which takes ~3 minutes, so it's not practical)

There's currently no way to run training (module-by-module) via lld (looked into it, there's potentially one way, but it's broken), so the command to re-run lld is effectively useless. We need a command to run clang, to mimic how lld sets up the optimization pipeline, and that's left to the user to decide, since there's too many options to parse in the lld command.  Maybe in the future we can figure out what a "good approximation" is based solely off the lld command, but even this is subject to change at any time (when someone adds new features for example)

compiler_opt/rl/corpus.py
Northbadge: ah well, just moved it to constant.py

PR #67, opened 2022-07-26
Minor fixes

Northbadge:
- ensure the '-cc1' flag exists for commands extracted from .cmd files, as it is required
- disabled yapf formatting on data_reader's dataset builder
- gracefully handle when a compile_command doesn't have the -o option
- check-license.sh now ignores shebang lines

### DIFF
--- a/check-license.sh
+++ b/check-license.sh
@@ -1,3 +1,4 @@
+#!/bin/bash
 # Copyright 2020 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
@@ -12,27 +13,33 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-#!/bin/bash
 
 exitcode=0
-for fname in `find . -name "*.py"`; 
-do
-  diff -q <(head -14 $fname) <(cat license-header.txt) > /dev/null
-  if [ $? -ne 0 ]
+
+check_license () {
+  HEADER=$(head -1 $1)
+  if [[ $HEADER == \#!* ]] # Ignore shebang line
   then
-    echo "$fname does not have license header. Please copy and paste ./license-header.txt"
+    HEADER=$(head -$3 $1 | sed 1d)
+  else
+    HEADER=$(head -$(($3 - 1)) $fname)
+  fi
+  if [[ "$HEADER" != "$2" ]]
+  then
+    echo "$1 does not have license header. Please copy and paste ./license-header.txt"
     exitcode=1
   fi
+}
+
+for fname in `find . -name "*.py"`;
+do
+  check_license $fname "$(cat license-header.txt)" 15
 done
 
-for fname in `find . -name "*.sh"`; 
+for fname in `find . -name "*.sh"`;
 do
-  diff -q <(head -13 $fname) <(tail -13 license-header.txt) > /dev/null
-  if [ $? -ne 0 ]
-  then
-    echo "$fname does not have license header. Please copy and paste ./license-header.txt"
-    exitcode=1
-  fi
+  check_license $fname "$(tail -13 license-header.txt)" 14
 done
 
 exit ${exitcode}
+
--- a/compiler_opt/rl/corpus.py
+++ b/compiler_opt/rl/corpus.py
@@ -125,4 +125,8 @@ def _load_and_parse_command(
 
   cmdline.extend(additional_flags)
 
+  # The options read from a .cmd file must be run with -cc1
+  if not cmd_override and cmdline[0] != '-cc1':
+    cmdline = ['-cc1'] + cmdline
+
   return cmdline
--- a/compiler_opt/rl/corpus.py
+++ b/compiler_opt/rl/corpus.py
@@ -127,6 +127,6 @@ def _load_and_parse_command(
 
   # The options read from a .cmd file must be run with -cc1
   if not cmd_override and cmdline[0] != '-cc1':
-    cmdline = ['-cc1'] + cmdline
+    raise ValueError('-cc1 flag not present in .cmd file.')
 
   return cmdline
--- a/compiler_opt/tools/extract_ir.py
+++ b/compiler_opt/tools/extract_ir.py
@@ -227,6 +227,8 @@ class TrainingIRExtractor:
               llvm_objcopy_path: Optional[str] = None,
               cmd_filter: Optional[str] = None,
               thinlto_build: Optional[str] = None) -> Optional[str]:
+    if self._obj_relative_path is None:
+      return None
     if thinlto_build == 'local':
       return self._extract_lld_artifacts()
     return self._extract_clang_artifacts(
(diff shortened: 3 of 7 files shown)


compiler_opt/rl/corpus.py
```diff
@@ -125,4 +125,8 @@ def _load_and_parse_command(
 
   cmdline.extend(additional_flags)
 
+  # The options read from a .cmd file must be run with -cc1
+  if not cmd_override and cmdline[0] != '-cc1':
```
yundiqian, line 129: are we attaching '-cc1' before your code change (i mean before you start the whole refactoring)? what made the difference that we don't do it before but do it now?

compiler_opt/tools/extract_ir.py
```diff
@@ -241,7 +243,11 @@ def convert_compile_command_to_objectfile(command: Dict[str, str],
   cmd = command['command']
 
   cmd_parts = cmd.split()
-  obj_index = cmd_parts.index('-o') + 1
+  try:
+    obj_index = cmd_parts.index('-o') + 1
+  except ValueError:
```
yundiqian, line 248: when does this happen?

Northbadge: looks like when there's a non-clang command, this is from Chrome
`
../../build/android/native_flags/argcapture.py  flagcapture/default_ccflags.txt  -DOFFICIAL_BUILD -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -D_FORTIFY_SOURCE=2 -D_GNU_SOURCE -DANDROID -DHAVE_SYS_UIO_H -DANDROID_NDK_VERSION_ROLL=r23_1 -D_LIBCPP_ABI_UNSTABLE -D_LIBCPP_DISABLE_VISIBILITY_ANNOTATIONS -D_LIBCXXABI_DISABLE_VISIBILITY_ANNOTATIONS -D_LIBCPP_ENABLE_NODISCARD -DCR_LIBCXX_REVISION=79a2e924d96e2fc1e4b937c42efd08898fa472d7 -DNDEBUG -DNVALGRIND -DDYNAMIC_ANNOTATIONS_ENABLED=0  -I../.. -Iflagcapture/gen -I../../buildtools/third_party/libc++  -fprofile-sample-use=../../chrome/android/profiles/afdo.prof -fprofile-sample-accurate -Wall -Werror -Wextra -Wimplicit-fallthrough -Wunreachable-code-aggressive -Wthread-safety -Wno-missing-field-initializers -Wno-unused-parameter -Wloop-analysis -Wno-unneeded-internal-declaration -Wenum-compare-conditional -Wno-psabi -Wno-ignored-pragma-optimize -Wno-unqualified-std-cast-call -Xclang -no-opaque-pointers -Wshadow -Xclang -fembed-bitcode=all -fno-delete-null-pointer-checks -fno-ident -fno-strict-aliasing --param=ssp-buffer-size=4 -fstack-protector -funwind-tables -fPIC -fcolor-diagnostics -fmerge-all-constants -fcrash-diagnostics-dir=../../tools/clang/crashreports -mllvm -instcombine-lower-dbg-declare=0 -ffp-contract=off -fcomplete-member-pointers -ffunction-sections -fno-short-enums --target=arm-linux-androideabi23 -march=armv7-a -mfloat-abi=softfp -mtune=generic-armv7-a -ffile-compilation-dir=. -no-canonical-prefixes -mfpu=neon -mthumb -Oz -fdata-sections -ffunction-sections -fno-unique-section-names -fomit-frame-pointer -gdwarf-4 -g1 -gdwarf-aranges -fdebug-info-for-profiling -fvisibility=hidden -Wheader-hygiene -Wstring-conversion -Wtautological-overlap-compare  -std=c++17 -Wno-trigraphs -fno-aligned-new -fno-exceptions -fno-rtti -nostdinc++ -isystem../../buildtools/third_party/libc++/trunk/include -isystem../../buildtools/third_party/libc++abi/trunk/include --sysroot=../../third_party/android_ndk/toolchains/llvm/prebuilt/linux-x86_64/sysroot -fvisibility-inlines-hidden
`

compiler_opt/rl/corpus.py
Northbadge: before my changes, -cc1 was assumed to be in the .cmd files, if I use `Debian clang version 13.0.1-3+build2`, -cc1 is included in the .cmd

If I use Chrome's version of clang, which is near-tip-of-tree llvm, there's no -cc1 in the .cmd file

Northbadge: seems like an llvm change probably, gonna go have a quick look

mtrofin: I'd leave the options "as-is" - in some cases, the build system runs clang as a driver (despite running it on a single module); in others, the build system would run -cc1. Just adding -cc1 isn't necessarily correct, either.

Northbadge: it's adding -cc1 iff we're reading off a .cmd file i.e. the .llvmcmd section, which must have been written by cc1-clang, which should've been called either by the clang-driver or directly, right?

anyway, turns out setting LLVM's DCMAKE_BUILD_TYPE to Debug eats the -cc1, somewhere, causing it to not end up in .llvmcmd

Northbadge: settled on raising an exception instead of supporting this

boomanaiden154: I just ran into this issue when testing out some stuff for the regalloc model. More specifically, the issue seems to occur when `LLVM_ENABLE_ASSERTIONS` is turned on, even when `CMAKE_BUILD_TYPE` is set to `Release`. I'm not sure what flags the chromium clang has enabled, but perhaps they have this turned on? Also seems like it's most likely a bug in clang/LLVM since toggling `LLVM_ENABLE_ASSERTIONS` shouldn't change behavior in this area. I'll try and look into this issue a little bit tonight while letting models train on the LLVM side.

boomanaiden154: Just opened https://reviews.llvm.org/D130620. Should resolve the behavior with chromium if they're enabling assertions and also fix the debug/release behavioral differences.

Northbadge: Yep, my Chromium toolchain builds have assertions on, thanks for looking into it & nice work on the fix!

compiler_opt/tools/extract_ir.py
yundiqian: so our intended behavior is to skip in this case?

Northbadge: yes

yundiqian: it probably makes more sense to have it return None in this case, and have remove None in line 261 to 264, then you don't need to add the None check in line 230/231 (from long-term code maintainance perspective, it's more difficult to associate the None check to TrainingIRExtractor(None, None, None))

yundiqian: and also add a comment about this is for filtering non-clang command, when it could happen, etc

check-license.sh
```diff
@@ -1,3 +1,4 @@
+#!/bin/bash
```
yundiqian, line 1: accidental mix?

yundiqian (approved):

check-license.sh
Northbadge: Intentional- #! lines are supposed to be on the first line so it can be run with ./abc.sh (it’ll run the file with whatever executable is on that first line)

yundiqian: oh, i mean do you accidentally mix this file in, cause i used to see it in a different PR, but whatever...

PR #68, opened 2022-07-27
Add presubmit script

Northbadge:
- Locally runs everything the GitHub CI used to run, and does so in parallel (except running yapf first)
- Makes GitHub CI run the presubmit script

[relevant xkcd](https://xkcd.com/1205/)

### DIFF
--- a/presubmit.py
+++ b/presubmit.py
@@ -1,0 +1,137 @@
+#!/usr/bin/env python3
+# coding=utf-8
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Pre-submit checks, equivalent to GitHub CI runs."""
+import functools
+import pathlib
+import signal
+import subprocess
+import sys
+import threading
+
+from absl import app
+from absl import flags
+from absl import logging
+
+FLAGS = flags.FLAGS
+
+_READ_ONLY = flags.DEFINE_boolean(
+    'read_only', False, 'Do not modify any files.', short_name='ro')
+# -v is used by absl
+_VERBOSE = flags.DEFINE_boolean(
+    'verbose', False, 'Print stdout.', short_name='vv')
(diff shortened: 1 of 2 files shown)


yundiqian (approved):
lgtm, just have mircea take a look at the shell script

check-license.sh
```diff
@@ -12,13 +13,19 @@
...
 # limitations under the License.
 
-#!/bin/bash
 
 exitcode=0
 for fname in `find . -name "*.py"`; 
 do
-  diff -q <(head -14 $fname) <(cat license-header.txt) > /dev/null
-  if [ $? -ne 0 ]
+  HEADER=$(head -1 $fname)
```
mtrofin, line 20: lgtm, but... if you're fixing this, mind also making this a function? :) (by "this" I mean the guts that check the header and all that)

presubmit.py
```diff
@@ -0,0 +1,128 @@
+#!/usr/bin/python3
```
mtrofin, line 1: how do we maintain this, i.e. say we modify the CI. I'd prefer not needing to modify it in 2 places.

is there a way to run the CI script locally that we can use instead?

check-license.sh
Northbadge: I’d vote for replacing it with a python script…

presubmit.py
Northbadge: Can’t run GH CI locally without an additional dependency(act) and docker it seems, but we could run this on GH CI instead

mtrofin: I just tried `act` and it's super easy: apt-get -ed docker and act, then `sudo act` in the root of this repo.

Let's not do the CI change, and instead use act, otherwise we add work for ourselves wrt having to maintain a CI-like replacement; we also need to figure out stuff like how to handle error output - are we ok to leave it interleaved? (I don't think that'd be very helpful). With act + CI we don't have to worry about any of that.

Northbadge: The motivation for this was to have a fast, and local way of running the CI steps, for development. 

act is fine and all but it isn't fast since GH CI is limited to doing steps sequentially, which using a script that spawns its own processes solves. (1m37s vs 32s, without act redownloading packages) So I still think we should have the script, even if using act.

I'm not sure what you mean by interleaved output though, you can see what the output looks with -vvv here https://github.com/Northbadge/ml-compiler-opt/runs/7550324734?check_suite_focus=true having it fail-fast on the CI is also an option instead of having it continue running the other jobs, which replicates what the CI does.

boomanaiden154: If we refactor the Github CI to use multiple jobs (encapsulating each current step in a separate job), that should allow both the Github CI and `act` to run things in parallel. We'll probably be spending a little bit more time in setup (I would imagine even more on the Github CI side), but either way things should be significantly faster and all the steps that would need to be performed locally a pretty cheap (mostly just container creation/maybe copying some files).

PR #72, opened 2022-08-01
Fix metric logging

Northbadge:
- Decouples the summary intervals, so one is not reliant on the other

- Switches to modulo instead of remembering the last step number- this is more robust/less-error-prone, especially if training is restarted at some non-zero step N (which previously caused no scalar data to be written, at least for the first N + interval steps)


yundiqian:
thanks for the fix!

PR #75, opened 2022-08-01
Set max workers for LocalWorkerManager

Northbadge:
- Prevents attempting to spawn num_workers * min(32, cores) number of clangs (python errors out)
- max_worker count is adjustable via `pool_threads` parameter in LocalWorkerPool constructor

### DIFF
--- a/compiler_opt/distributed/local/local_worker_manager.py
+++ b/compiler_opt/distributed/local/local_worker_manager.py
@@ -35,6 +35,7 @@ import multiprocessing.connection
 import queue  # pylint: disable=unused-import
 import threading
 
+import gin
 from absl import logging
 # pylint: disable=unused-import
 from compiler_opt.distributed.worker import Worker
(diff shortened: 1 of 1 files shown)


compiler_opt/distributed/local/local_worker_manager.py
```diff
@@ -208,16 +217,25 @@ def __dir__(self):
...
+               count: Optional[int],
+               *args,
+               pool_threads: int = 1,
                **kwargs):
     if not count:
       count = multiprocessing.cpu_count()
     self._stubs = [
-        _make_stub(worker_class, *args, **kwargs) for _ in range(count)
+        _make_stub(worker_class, pool_threads, *args, **kwargs)
+        for _ in range(count // pool_threads)
```
mtrofin, line 234: nit: for _ in range(max(1, count//pool_threads))

... or logging.error if that nr is 0?

mtrofin: a, nevermind - saw the code below

PR #76, opened 2022-08-01
import modules instead of importing packages

yundiqian:


mtrofin:
Awesome - this diverged earlier in pytype when we were doing `--protocols`, happy we don't need to anymore (now that we dropped that flag)

PR #77, opened 2022-08-01
Add Corpus abstraction & sorting w/ size

Northbadge:
- Rather than pass around a list of module_specs, pass around a Corpus object instead
- Corpus has built in filter and sample method
- sample will sort by size descending, with the goal of optimizing compile order for time

### DIFF
--- a/compiler_opt/rl/corpus.py
+++ b/compiler_opt/rl/corpus.py
@@ -13,7 +13,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """ModuleSpec definition and utility command line parsing functions."""
-import itertools
 import random
 import re
 
@@ -53,12 +52,12 @@ class Corpus:
   @classmethod
   def from_module_specs(cls, module_specs: List[ModuleSpec]):
     """Construct a Corpus from module specs. Mostly for testing purposes."""
-    corp = cls.__new__(cls)  # Avoid calling __init__
-    super(cls, corp).__init__()
-    corp._module_specs = list(module_specs)  # Don't mutate the original list.
-    corp._module_specs.sort(key=lambda m: m.size, reverse=True)
-    corp.root_dir = None
-    return corp
+    cps = cls.__new__(cls)  # Avoid calling __init__
+    super(cls, cps).__init__()
+    cps._module_specs = list(module_specs)  # Don't mutate the original list.
+    cps._module_specs.sort(key=lambda m: m.size, reverse=True)
+    cps.root_dir = None
+    return cps
 
   def sample(self,
              k: int,
--- a/compiler_opt/rl/corpus.py
+++ b/compiler_opt/rl/corpus.py
@@ -50,8 +50,7 @@ class Corpus:
           data_path=data_path,
           additional_flags=additional_flags,
           delete_flags=delete_flags)
-    self._root_dir = data_path
-    self.size = len(self._module_specs)
+    self.root_dir = data_path
     self._module_specs.sort(key=lambda m: m.size, reverse=True)
 
   def sample(self, k: int) -> List[ModuleSpec]:
@@ -68,6 +67,9 @@ class Corpus:
   def filter(self, p: re.Pattern):
     """Filters module specs, keeping those which match the provided pattern."""
     self._module_specs = [ms for ms in self._module_specs if p.match(ms.name)]
+
+  def __len__(self):
+    return len(self._module_specs)
 
 
 def _build_modulespecs_from_datapath(
--- a/compiler_opt/rl/corpus.py
+++ b/compiler_opt/rl/corpus.py
@@ -13,6 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """ModuleSpec definition and utility command line parsing functions."""
+import math
 import random
 import re
 
@@ -33,6 +34,51 @@ class ModuleSpec:
   name: str
   exec_cmd: Tuple[str, ...] = ()
   size: int = 0
+
+
+class SamplerBucketRoundRobin:
+  """Calls return a list of module_specs sampled randomly from n buckets, in
+  randomized-round-robin order. The buckets are sequential sections of
+  module_specs of roughly equal lengths."""
+
+  def __init__(self):
+    self._ranges = {}
+
+  def __call__(self,
+               module_specs: List[ModuleSpec],
+               k: int,
+               n: int = 20) -> List[ModuleSpec]:
+    """
+    Args:
+      module_specs: list of module_specs to sample from
+      k: number of modules to sample
+      n: number of buckets to use
+    """
+    # Credits to yundi@ for the highly optimized algo
+    # Essentially, split module_specs into k buckets, then define the order of
+    # visiting the k buckets such that it approximates the behaviour of having
+    # n buckets.
+    specs_len = len(module_specs)
+    if (specs_len, k, n) not in self._ranges:
+      quotient, remainder = divmod(k, n)
+      # rev_map maps from bucket # (implicitly via index) to order of visiting.
+      # lower values should be visited first, and earlier indices before later.
+      rev_map = list(range(quotient)) * n + list(range(remainder))
+      # mapping defines the order in which buckets should be visited.
+      mapping = [t[0] for t in sorted(enumerate(rev_map), key=lambda x: x[1])]
+
+      # generate the buckets ranges, in the order which they should be visited.
+      bucket_size_float = specs_len / k
+      self._ranges[(specs_len, k, n)] = tuple(
+          (math.floor(bucket_size_float * i),
+           math.floor(bucket_size_float * (i + 1))) for i in mapping)
+
+    rand_between = random.randrange
+
+    return [
+        module_specs[rand_between(start, end)]
+        for start, end in self._ranges[(specs_len, k, n)]
+    ]
 
 
 class Corpus:
@@ -62,8 +108,8 @@ class Corpus:
   def sample(self,
              k: int,
              sort: bool = False,
-             sampler=random.sample) -> List[ModuleSpec]:
-    """Samples `k` module_specs, sorting by size descending."""
+             sampler=SamplerBucketRoundRobin()) -> List[ModuleSpec]:
+    """Samples `k` module_specs, optionally sorting by size descending."""
     k = min(len(self._module_specs), k)
     if k < 1:
       raise ValueError('Attempting to sample <1 module specs from corpus.')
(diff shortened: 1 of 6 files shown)


compiler_opt/rl/local_data_collector.py
```diff
@@ -34,7 +33,7 @@ class LocalDataCollector(data_collector.DataCollector):
 
   def __init__(
       self,
-      module_specs: List[corpus.ModuleSpec],
+      corp: corpus.Corpus,
```
mtrofin, line 36: corpus

Northbadge: Yep, conflicts with the package. but corpora is technically plural

compiler_opt/rl/corpus.py
```diff
@@ -65,14 +103,17 @@ def build_modulespecs_from_datapath(
...
+    full_module_path = os.path.join(data_path, rel_module_path)
     exec_cmd = _load_and_parse_command(
-        module_path=os.path.join(data_path, module_path),
+        module_path=full_module_path,
         has_thinlto=has_thinlto,
         additional_flags=additional_flags,
         delete_flags=delete_flags,
         cmd_override=cmd_override)
-    module_specs.append(ModuleSpec(name=module_path, exec_cmd=tuple(exec_cmd)))
+    size = os.path.getsize(full_module_path + '.bc')
```
mtrofin, line 116: does this change the estimate on line 105?

Northbadge: Nope. But now that you've mentioned it the estimate on 105 does have the advantage of all the files being cached in RAM...

compiler_opt/rl/corpus.py
```diff
@@ -30,9 +32,45 @@ class ModuleSpec:
...
+    self.size = len(self._module_specs)
+    self._module_specs.sort(key=lambda m: m.size, reverse=True)
+
+  def sample(self, k: int) -> List[ModuleSpec]:
+    """Samples `k` module_specs, sorting by size descending."""
+    k = min(len(self._module_specs), k)
+    if k < 1:
+      raise ValueError('Attempting to sample <1 module specs from corpus.')
+    return list(
+        sorted(
```
mtrofin, line 62: this is a breaking change tho, no?

but - IIRC, the goal of starting to look at the file size was to enable creating samples where the distribution of the file size in the sample doesn't vary by much - the intuition being that .bc size is a reasonable proxy for compile time, and if samples always have a few gigantic modules, they'll always be ignored.

I don't think the change achieves that, right?

Separately, I'd offer the size-based sampling as an opt-in.

Northbadge: > this is a breaking change tho, no?

breaking in the sense that it changes compile order, yes.

> creating samples where the distribution of the file size in the sample doesn't vary by much

I wasn't aware of this, was it intentional to ignore large modules? Do we want to modify the distribution of the sample in some way?

I'll make the sorting optional, then it'll have the same effect as before on training. The sorting was more for the validation data collector (which was consistently being unlucky with a large module at the back)

mtrofin: > I wasn't aware of this, was it intentional to ignore large modules? Do we want to modify the distribution of the sample in some way?

Not intentional. So, the data collector is happy to fire N jobs and get a subset back and cancel the rest, in the interest of time. To echo @yundiqian , consider this: say you have N=100. You consistently get 1 module that's enormous. That one would take 5 minutes to compile. Everything else takes 5 seconds.

The "size-based sampling" idea was about having a way we can say "we prefer having similarly sized modules in each sample, if possible". Not sure how this would be formally expressed though.

What's the validation data collector?

compiler_opt/rl/local_data_collector.py
Northbadge: there are several files both importing corpus and using a Corpus; only doing the renamed import in this file is probably gonna be more confusing, renaming everything might be better?

compiler_opt/rl/corpus.py
Northbadge: > "we prefer having similarly sized modules in each sample, if possible"

does "Split the modules by size into N buckets, based on size, in size-order, such that the smallest bucket has the smallest 1/N modules. Sample round-robin from the N buckets, schedule tasks in the order of sampling. " sound about right? Minimizes risk of only having large modules compile first and leaving no time left for tiny modules.

> What's the validation data collector?

Compiles the entirety of a separate-from-training "validation" dataset and returns the performance of the currently trained model on it as the model is being trained. The thing we talked about offline last week

mtrofin: the bucketing would work, but not sure if it's sufficiently randomized - @yundiqian?

> validation data collector
Aaa! OK. Got it. We could have that guy do the sorting on its side, or? Anyway, for now, let's do what you have as an opt-in (unless the validation one can do it on its end - haven't seen the impl).

compiler_opt/rl/corpus.py
mtrofin: you mean all the file paths? but so are the `ModuleSpec`s, no? 

On that - separately - we could look at making `ModuleSpec` use `__slots__` to reduce total memory footprint of a collection of these things.

compiler_opt/rl/corpus.py
yundiqian: this definitely breaks the randomization, but i'm not sure whether it affects the result or not. we can implement it as the additional sampling algorithm so both the random sampling and bucketing sampling are supported

btw, if we do the ordering of the sampled modules here, for the round robin dispatching algorithm, we should probably do 1 2 3 3 2 1 instead of 1 2 3 1 2 3 in dispatching

compiler_opt/rl/corpus.py
```diff
@@ -30,9 +32,47 @@ class ModuleSpec:
...
   name: str
   exec_cmd: Tuple[str, ...] = ()
+  size: int = 0
 
 
-def build_modulespecs_from_datapath(
+class Corpus:
+  """Represents a corpus. Comes along with some utility functions."""
+
+  def __init__(self,
```
yundiqian, line 41: only support one way of initialization?

compiler_opt/rl/corpus.py
```diff
@@ -30,9 +32,47 @@ class ModuleSpec:
...
+               module_specs: Optional[List[ModuleSpec]] = None):
+    if module_specs is not None:
+      self._module_specs = module_specs
+    else:
+      self._module_specs = _build_modulespecs_from_datapath(
+          data_path=data_path,
+          additional_flags=additional_flags,
+          delete_flags=delete_flags)
+    self.root_dir = data_path
+    self._module_specs.sort(key=lambda m: m.size, reverse=True)
```
yundiqian, line 54: why do sorting here?

compiler_opt/rl/corpus.py
```diff
@@ -30,9 +32,47 @@ class ModuleSpec:
...
+
+  def sample(self, k: int) -> List[ModuleSpec]:
+    """Samples `k` module_specs, sorting by size descending."""
+    k = min(len(self._module_specs), k)
+    if k < 1:
+      raise ValueError('Attempting to sample <1 module specs from corpus.')
+    return list(
```
yundiqian, line 61: nit: can you write sampling with one line of code, and sorting with another line of code to emphasize that the sorting does not affect the sampling result?

compiler_opt/rl/corpus.py
Northbadge: was an artifact from development, but it's useful for the bucketing as well so will be left as-is

compiler_opt/rl/corpus.py
Northbadge: on the 1 2 3 3 2 1 part: you mean to pick the buckets in random order, right?

compiler_opt/rl/corpus.py
Northbadge: do you mean adding a way for composing corpora together?

compiler_opt/rl/corpus.py
Northbadge: I meant that the file/directory is cached to RAM, so accesses were faster when I got that 7s/30k modules number, don't think the size of the ModuleSpecs would be concerning (for a single project's worth of ModuleSpecs)

compiler_opt/rl/corpus.py
yundiqian: no, i mean doing the 12345543211234554321 in dispatching assuming 20 jobs for 5 workers if the 20 jobs are sorted with descending order in compilation time. this gives you slightly more balanced workload for workers

if you do the current 12345123451234512345, the worker 1 is of the heaviest load and worker 5 is of the lightest load.

compiler_opt/rl/corpus.py
yundiqian: i mean currently the __init__ is a bit confusing cause you are allowing both passing in a modulespec list or passing in data_path, additional_flag, delete_flag to construct the modulespec. this interface is confusing to users. you can either: 1) only support init from one way; or 2) have separate init for these two ways

compiler_opt/rl/corpus.py
Northbadge: ah, that might be better fixed with a load-balancing job dispatcher though, since there isn't a guarantee that 123 321 123 is the most optimal scheduling. Right now I have it doing 123 321 213 231, which should be "even" enough (on average, anyway) for the round-robin job dispatcher.

compiler_opt/rl/corpus.py
```diff
@@ -30,9 +33,74 @@ class ModuleSpec:
...
+
+  def filter(self, p: re.Pattern):
+    """Filters module specs, keeping those which match the provided pattern."""
+    self._module_specs = [ms for ms in self._module_specs if p.match(ms.name)]
+
+  def __len__(self):
+    return len(self._module_specs)
+
+
+def sampler_bucket_round_robin(module_specs: List[ModuleSpec],
```
yundiqian, line 84: the current logic is quite complicated, why not directly do sampling?

Northbadge:
might be worth discussing whether we want to have train_locally to default to the bucket sampler (later today?)

compiler_opt/rl/corpus.py
yundiqian: i get a bit confused of the function, originally i thought you want to sample k modules that are of similar size so that the time to compile them was similar so that the data collection can be more efficient

it seems that what you are doing here is something reverse --- have a more spread distribution of compilation time, i'm wondering what's the benefit of it compared with just a random sampling?

compiler_opt/rl/corpus.py
```diff
@@ -30,9 +33,100 @@ class ModuleSpec:
...
+    # Credits to yundi@ for the highly optimized algo
+    # Essentially, split module_specs into k buckets, then define the order of
+    # visiting the k buckets such that it approximates the behaviour of having
+    # n buckets.
+    specs_len = len(module_specs)
+    if (specs_len, k, n) not in self._ranges:
+      quotient, remainder = divmod(k, n)
+      # rev_map maps from bucket # (implicitly via index) to order of visiting.
+      # lower values should be visited first, and earlier indices before later.
+      rev_map = list(range(quotient)) * n + list(range(remainder))
```
yundiqian, line 66: rev_map = [x % quotient for x in range(k)] for a more even break --- see k = 119, n = 20 for example

compiler_opt/tools/generate_default_trace.py
```diff
@@ -139,20 +138,19 @@ def main(_):
...
+    cps.filter(m)
 
   # Sampling if needed.
-  if _SAMPLING_RATE.value < 1:
-    sampled_modules = int(len(module_specs) * _SAMPLING_RATE.value)
-    module_specs = random.sample(module_specs, k=sampled_modules)
+  sampled_modules = int(len(cps) * _SAMPLING_RATE.value)
+  module_specs = cps.sample(k=sampled_modules, sort=True)
 
   # sort files by size, to process the large files upfront, hopefully while
```
yundiqian, line 155: delele sorting here

compiler_opt/tools/generate_default_trace.py
```diff
@@ -139,20 +138,19 @@ def main(_):
...
   if _MODULE_FILTER.value:
     m = re.compile(_MODULE_FILTER.value)
-    module_specs = [ms for ms in module_specs if m.match(ms.name)]
+    cps.filter(m)
 
   # Sampling if needed.
-  if _SAMPLING_RATE.value < 1:
-    sampled_modules = int(len(module_specs) * _SAMPLING_RATE.value)
-    module_specs = random.sample(module_specs, k=sampled_modules)
+  sampled_modules = int(len(cps) * _SAMPLING_RATE.value)
```
yundiqian, line 152: add a unittest in corpus_test.py for the k == len(cps) case

compiler_opt/rl/corpus.py
Northbadge: added a check for when k < n too

compiler_opt/tools/generate_default_trace.py
Northbadge: I had "test_bucket_sample_all", or did you mean to add another/something different?

yundiqian: yeah, this is what i'm looking for, sorry i missed it before

yundiqian (approved):

mtrofin:
lgtm

PR #78, opened 2022-08-02
Add development mode dockerfile and info on use

boomanaiden154:
Development mode dockerfile and instructions on use in a `README.md` inside a newly created `docker/` directory. Should be something useful for anyone working on the project who is struggling with creating a working build environment and also general development workflows that use docker.

### DIFF
--- a/experimental/docker/README.md
+++ b/experimental/docker/README.md
@@ -1,0 +1,23 @@
+# MLGO Development Dockerfiles
+
+This folder contains dockerfiles with all the dependencies necessary to do
+development work with MLGO. Note that these dockerfiles do not contain a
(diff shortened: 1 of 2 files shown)


docker/README.md
```diff
@@ -0,0 +1,23 @@
+# MLGO Development Dockerfiles
```
mtrofin, line 1: a thought: should we name the directory `experimental`, and then there's even more hinting that this is... well... experimental/best effort support?

boomanaiden154: Sounds like a decent idea to me. I moved the entire `docker/` directory to `experimental/docker/` in case we want to put anything else in there later and also to segment it a little bit better. Thanks for the suggestion.

PR #81, opened 2022-08-02
LocalWorker: Fix nr threads to 1

mtrofin:
From discussion on PR #75

### DIFF
--- a/compiler_opt/distributed/local/local_worker_manager.py
+++ b/compiler_opt/distributed/local/local_worker_manager.py
@@ -35,7 +35,6 @@ import multiprocessing.connection
 import queue  # pylint: disable=unused-import
 import threading
 
-import gin
 from absl import logging
 # pylint: disable=unused-import
 from compiler_opt.distributed.worker import Worker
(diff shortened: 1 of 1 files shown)


compiler_opt/distributed/local/local_worker_manager.py
```diff
@@ -120,13 +120,16 @@ def __init__(self):
...
+      # to handle high priority requests. The expectation is that the user
+      # achieves concurrency through multiprocessing, not multithreading.
       self._process = multiprocessing.Process(
           target=functools.partial(
               _run,
               worker_class=cls,
               in_q=self._send,
               out_q=self._receive,
-              threads=pool_threads,
+              threads=1,
```
yundiqian, line 132: just have max_workers=1 in line 76?

yundiqian (commented):
delete `import gin`

compiler_opt/distributed/local/local_worker_manager.py
mtrofin: yup

boomanaiden154:
I've found that setting the number of threads in the `ThreadPoolExecutor` to something greater than 1 helps with increasing CPU utilization, but that might only be when setting the `num_workers` CLI option to something significantly less than the number of threads available on the system. If set equal to the number of threads in the system though (the default), it is fine if I recall correctly. If it becomes an issue I can always just do some benchmarking and adjust the variable later.

PR #82, opened 2022-08-03
Enable building tflite on 'dev*'  buildbots

mtrofin:
Also bumping Ruy hash to capture last build-related fix

### DIFF
--- a/buildbot/buildbot_init.sh
+++ b/buildbot/buildbot_init.sh
@@ -97,6 +97,20 @@ userdel buildbot
 groupadd buildbot
 useradd buildbot -g buildbot -m -d /var/lib/buildbot
 
+if [[ "${HOSTNAME}" == ml-opt-dev* ]]
+then
+  echo "Building TFLite"
+  curdir="${PWD}"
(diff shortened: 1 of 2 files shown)


buildbot/buildbot_init.sh
```diff
@@ -97,6 +97,20 @@ userdel buildbot
 groupadd buildbot
 useradd buildbot -g buildbot -m -d /var/lib/buildbot
 
+if [[ "${HOSTNAME}" == ml-opt-dev* ]]
+then
+  echo "Building TFLite"
+  curdir="${PWD}"
```
petrhosek, line 103: It might be cleaner to use `pushd /tmp/tflitebuild` instead  of `curdir="${PWD}"` followed by `cd /tmp/tflitebuild` and later do `popd` instead of `cd "${curdir}"`.

petrhosek (approved):

buildbot/buildbot_init.sh
mtrofin: ah yes! pushd. And I can pushd /tmp/tflitebuild / popd

PR #83, opened 2022-08-04
Fix worker shutdown

Northbadge:
If process was killed while its output queue is accessed, the queue would be left broken, causing msg_pump to not die, and .join() to never unblock.

Also, fix a misc. indentation bug on a test case

### DIFF
--- a/compiler_opt/distributed/local/local_worker_manager.py
+++ b/compiler_opt/distributed/local/local_worker_manager.py
@@ -204,8 +204,8 @@ def _make_stub(cls: 'type[Worker]', *args, **kwargs):
 
     def shutdown(self):
       try:
-        # Killing the process triggers observer death, which triggers msg_pump
-        # death
+        # Killing the process triggers observer exit, which triggers msg_pump
+        # exit
         self._process.kill()
       except:  # pylint: disable=bare-except
         pass
--- a/compiler_opt/distributed/local/local_worker_manager.py
+++ b/compiler_opt/distributed/local/local_worker_manager.py
@@ -59,14 +59,10 @@ class TaskResult:
   value: Any
 
 
-def _run_impl(in_q: 'queue.Queue[Task]', out_q: 'queue.Queue[TaskResult]',
+def _run_impl(pipe: multiprocessing.connection.Connection,
               worker_class: 'type[Worker]', *args, **kwargs):
   """Worker process entrypoint."""
-  # Note: the out_q is typed as taking only TaskResult objects, not
-  # Optional[TaskResult], despite that being the type it is used on the Stub
-  # side. This is because the `None` value is only injected by the Stub itself.
 
-  # `threads` is defaulted to 1 in LocalWorkerPool's constructor.
   # A setting of 1 does not inhibit the while loop below from running since
   # that runs on the main thread of the process. Urgent tasks will still
   # process near-immediately. `threads` only controls how many threads are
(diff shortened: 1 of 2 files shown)


Northbadge:
It just occurred to me that if the child process was killed externally while it was accessing a queue, the same bug would resurface... Might need to put msg_pump on a different death mechanism (i.e. not a queue)

yundiqian (approved):

compiler_opt/distributed/local/local_worker_manager.py
```diff
@@ -128,8 +128,12 @@ def __init__(self):
       # thread drainig the receive queue
       self._pump = threading.Thread(target=self._msg_pump)
 
+      # Set the state of this worker to "dead" if the process dies naturally.
```
mtrofin, line 131: "Naturally" meaning because it raised an exception, because that's the only reason it'd happen, correct?

Northbadge: Correct

mtrofin:
> It just occurred to me that if the child process was killed externally while it was accessing a queue, the same bug would resurface... Might need to put msg_pump on a different death mechanism (i.e. not a queue)

Yup. I think the root of the problem is the state of the reader lock on the queue buffer. We may be better off working directly with `multiprocessing.connection.Connection` objects from 2 non-duplex `Pipe`s - wdyt?

Northbadge:
> > It just occurred to me that if the child process was killed externally while it was accessing a queue, the same bug would resurface... Might need to put msg_pump on a different death mechanism (i.e. not a queue)
> 
> Yup. I think the root of the problem is the state of the reader lock on the queue buffer. We may be better off working directly with `multiprocessing.connection.Connection` objects - wdyt?

Switched to `Pipe`s now, which use `Connection`s

compiler_opt/distributed/local/local_worker_manager.py
```diff
@@ -189,20 +189,23 @@ def remote_call(*args, **kwargs):
...
+                      is_urgent=cls.is_priority_method(name)))
             self._map[msgid] = result_future
         return result_future
 
       return remote_call
 
     def shutdown(self):
       try:
+        # Killing the process triggers observer death, which triggers msg_pump
+        # death
```
mtrofin, line 208: nit: msg_pump _exit_ (because 'death' may be taken as forceful exit, but that's not the case)

compiler_opt/distributed/local/local_worker_manager.py
```diff
@@ -114,33 +117,30 @@ class _Stub():
     """Client stub to a worker hosted by a process."""
 
     def __init__(self):
-      self._send: 'queue.Queue[Task]' = multiprocessing.get_context().Queue()
-      self._receive: 'queue.Queue[Optional[TaskResult]]' = \
-        multiprocessing.get_context().Queue()
+      parent_pipe, child_pipe = multiprocessing.get_context().Pipe()
```
mtrofin, line 120: nit: can you call them incoming_pipe, outgoing_pipe? or task_result_pipe / task_pipe? (just to clarify more from the name what objects travel the pipes, which should also clarify what they are used for)

mtrofin:
lgtm, just 2 nits.

compiler_opt/distributed/local/local_worker_manager.py
Northbadge: the two vars don't exactly represent an incoming/outgoing pipe, there's just "one", bidirectional pipe. The two vars are the two ends of the pipe, .send from one corresponds to .recv on the other

mtrofin: yes, but one is traversed by TaskResults, the other by Tasks, no?

Northbadge: The underlying unidirectional connections, yes. 
The connections returned by .Pipe(), no- 
`parent_pipe.send(Task)`
`Optional[TaskResult] = parent_pipe.recv()`
 `child_pipe.send(TaskResult)`
 `Task = child_pipe.recv()`

Northbadge: I can see how `_pipe` might be a confusing suffix though, but {incoming|outgoing|task|task_result}_pipe would be even more counter-intuitive (like `incoming_pipe.send(Task)`, `Optional[TaskResult] = task_pipe.recv()`)

The python docs just used conn1, conn2

mtrofin: sorry, I probably caused the confusion with giving 4 naming choices. Let's say just: `task_pipe, task_result_pipe`. Would that cause confusion?

Northbadge: still yes- task_pipe.send(Task), TaskResult = task_pipe.recv(); task_result_pipe.send(TaskResult) Task=task_result_pipe.recv(),

mtrofin: ooo! I see. OK. lgtm then

PR #84, opened 2022-08-05
Refactor generate_default_trace to accept build command CLI modification

boomanaiden154:
options from gin_configs

Currently there are two places where we are specifying
deletions/additional flags when processing commands for building object
files, in ppo_nn_agent.gin for both the inlining and regalloc case. This
patch refactors generate_default_trace.py into two separate files so
that the actual function that handles the tracing can be configured in a
gin file. There are new gin files for the regalloc and inliner case that
specify options specifically for the default trace. This makes
maintenance easier as if these flag options need to change, they only
have to be changed in one place (per optimization case). This also makes
it possible to still warmstart models easily if the flags between the
two cases diverge, and it makes it possible to specify other options for
the default trace at the gin config level. This also makes it easier to
enable development mode features such as those added in at https://reviews.llvm.org/D131209
(gated behind a LLVM flag).

### DIFF
--- a/compiler_opt/rl/inlining/gin_configs/ppo_nn_agent.gin
+++ b/compiler_opt/rl/inlining/gin_configs/ppo_nn_agent.gin
@@ -20,18 +20,6 @@ train_eval.deploy_policy_name='saved_collect_policy'
 train_eval.use_random_network_distillation=False
 train_eval.moving_average_decay_rate=0.8
 
-# TODO(b/233935329): The following clang flags need to be tied to a corpus
-# rather than to a training tool invocation.
-
-# List of flags to add to clang compilation command. The flag names should
-# match the actual flags provided to clang. An example for AFDO reinjection:
-# train_eval.flags_to_add=('-fprofile-sample-use=/path/to/gwp.afdo','-fprofile-remapping-file=/path/to/prof_remap.txt')
-train_eval.additional_compilation_flags=()
-
-# List of flags to remove from clang compilation command. The flag names 
-# should match the actual flags provided to clang.'
-train_eval.delete_compilation_flags=('-split-dwarf-file','-split-dwarf-output','-fthinlto-index','-fprofile-sample-use','-fprofile-remapping-file')
-
 # RandomNetworkDistillation configs, off if train_eval.use_random_network_distillation=False.
 RandomNetworkDistillation.encoding_network = @encoding_network.EncodingNetwork
 RandomNetworkDistillation.learning_rate = 1e-4
--- a/compiler_opt/rl/problem_configuration.py
+++ b/compiler_opt/rl/problem_configuration.py
@@ -99,3 +99,19 @@ class ProblemConfiguration(metaclass=abc.ABCMeta):
   @abc.abstractmethod
   def get_runner_type(self) -> 'type[compilation_runner.CompilationRunner]':
     raise NotImplementedError
+
+  # TODO(b/233935329): The following clang flags need to be tied to a corpus
+  # rather than to a training tool invocation.
+
+  # List of flags to add to clang compilation command. The flag names should
+  # match the actual flags provided to clang. An example for AFDO reinjection:
+  # return ['-fprofile-sample-use=/path/to/gwp.afdo',
+  #  '-fprofile-remapping-file=/path/to/prof_remap.txt']
+  def flags_to_add(self) -> Tuple[str, ...]:
+    return ()
+
+  # List of flags to remove from clang compilation command. The flag names
+  # should match the actual flags provided to clang.'
+  def flags_to_delete(self) -> Tuple[str, ...]:
+    return ('-split-dwarf-file', '-split-dwarf-output', '-fthinlto-index',
+            '-fprofile-sample-use', '-fprofile-remapping-file')
--- a/compiler_opt/tools/generate_default_trace.py
+++ b/compiler_opt/tools/generate_default_trace.py
@@ -136,12 +136,13 @@ def main(_):
       _GIN_FILES.value, bindings=_GIN_BINDINGS.value, skip_unknown=False)
   logging.info(gin.config_str())
 
+  config = registry.get_configuration()
+
   logging.info('Loading module specs from corpus.')
   module_specs = corpus.build_modulespecs_from_datapath(
       _DATA_PATH.value,
-      delete_flags=('-split-dwarf-file', '-split-dwarf-output',
-                    '-fthinlto-index', '-fprofile-sample-use',
-                    '-fprofile-remapping-file'))
+      additional_flags=config.flags_to_add(),
+      delete_flags=config.flags_to_delete())
   logging.info('Done loading module specs from corpus.')
 
   if _MODULE_FILTER.value:
--- a/compiler_opt/tools/generate_default_trace_test.py
+++ b/compiler_opt/tools/generate_default_trace_test.py
@@ -60,9 +60,16 @@ class MockCompilationRunner(compilation_runner.CompilationRunner):
 
 
 class GenerateDefaultTraceTest(absltest.TestCase):
+  def setUp(self):
+    with gin.unlock_config():
+      gin.parse_config_files_and_bindings(
+          config_files=['compiler_opt/rl/inlining/gin_configs/common.gin'],
+          bindings=None)
+    return super().setUp()
 
   @mock.patch('compiler_opt.tools.generate_default_trace.get_runner')
   def test_api(self, mock_get_runner):
+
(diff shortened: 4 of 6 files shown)


yundiqian:
I think the change proposed here is a bit over-engineering for this problem, and actually creates more complexity from maintanance perspective. 

An alternative way is to add it as a method in {Inlining|RegAlloc}Config class, and have it passed through gin and return there.

mtrofin:
side- comment, because I'm seeing a few patches about compilation flags - we still need to make the change from just "add/delete" to "add/replace/delete". "Replace" - which happens e.g. for profile file locations - is better modeled explicitly because it allows catching setup errors where, say, a profile was meant to be used during compilation, but the experimenter forgets to pass one. Removing and not adding the flag _will_ work - compilation will just chug along - but the results will be garbage, and hard to tell why.

Opened Issue #85

boomanaiden154:
I just rewrote the patch to add methods to the problem configuration class that can easily be overwritten in a specific problem case. Definitely a much cleaner patch that looks like it should be easier maintenance wise in the future too.

compiler_opt/rl/problem_configuration.py
```diff
@@ -99,3 +99,21 @@ def get_nonnormalized_features(self) -> Iterable[str]:
   @abc.abstractmethod
   def get_runner_type(self) -> 'type[compilation_runner.CompilationRunner]':
     raise NotImplementedError
+
+  # TODO(b/233935329): The following clang flags need to be tied to a corpus
+  # rather than to a training tool invocation.
+
+  # List of flags to add to clang compilation command. The flag names should
+  # match the actual flags provided to clang. An example for AFDO reinjection:
+  # train_eval.additional_flags=('-fprofile-sample-use=/path/to/gwp.afdo','-fprofile-remapping-file=/path/to/prof_remap.txt')
```
yundiqian, line 108: update the comments here

compiler_opt/tools/generate_default_trace.py
```diff
@@ -136,12 +136,13 @@ def main(_):
...
 
+  config = registry.get_configuration()
+
   logging.info('Loading module specs from corpus.')
   module_specs = corpus.build_modulespecs_from_datapath(
       _DATA_PATH.value,
-      delete_flags=('-split-dwarf-file', '-split-dwarf-output',
-                    '-fthinlto-index', '-fprofile-sample-use',
-                    '-fprofile-remapping-file'))
+      delete_flags=config.flags_to_delete(),
```
yundiqian, line 144: nit: reverse the order

compiler_opt/rl/inlining/gin_configs/ppo_nn_agent.gin
```diff
@@ -20,18 +20,6 @@ train_eval.deploy_policy_name='saved_collect_policy'
...
-# rather than to a training tool invocation.
-
-# List of flags to add to clang compilation command. The flag names should
-# match the actual flags provided to clang. An example for AFDO reinjection:
-# train_eval.flags_to_add=('-fprofile-sample-use=/path/to/gwp.afdo','-fprofile-remapping-file=/path/to/prof_remap.txt')
-train_eval.additional_compilation_flags=()
-
-# List of flags to remove from clang compilation command. The flag names 
-# should match the actual flags provided to clang.'
-train_eval.delete_compilation_flags=('-split-dwarf-file','-split-dwarf-output','-fthinlto-index','-fprofile-sample-use','-fprofile-remapping-file')
```
kshiteejm, line 33: Removing these lines will break the google internal trainer (train_with_rpc.py) since this gin file is shared by train_locally.py and train_with_rpc.py. 

I think to avoid such unexpected conflicts we should have a separate internal-only copy of gin files. What do you think @yundiqian?

yundiqian: we don't need to, we can also do similar things internally, but yes, there needs to be a fix

compiler_opt/tools/generate_default_trace_test.py
```diff
@@ -59,10 +59,27 @@ def collect_data(self, module_spec, tf_policy_path, reward_stat):
...
+    return []
+
+
+def mock_get_configuration():
+  return MockProblemConfiguration()
+
+
 class GenerateDefaultTraceTest(absltest.TestCase):
 
+  @mock.patch('compiler_opt.rl.registry.get_configuration')
```
yundiqian, line 78: what is the problem if you don't have the mocking here?

compiler_opt/rl/problem_configuration.py
```diff
@@ -99,3 +99,21 @@ def get_nonnormalized_features(self) -> Iterable[str]:
...
     raise NotImplementedError
+
+  # TODO(b/233935329): The following clang flags need to be tied to a corpus
+  # rather than to a training tool invocation.
+
+  # List of flags to add to clang compilation command. The flag names should
+  # match the actual flags provided to clang. An example for AFDO reinjection:
+  # return ['-fprofile-sample-use=/path/to/gwp.afdo',
+  #  '-fprofile-remapping-file=/path/to/prof_remap.txt']
+  def flags_to_add(self) -> Iterable[str]:
```
yundiqian, line 110: can you still make them gin configurable?

compiler_opt/rl/inlining/gin_configs/ppo_nn_agent.gin
kshiteejm: agreed

compiler_opt/tools/generate_default_trace_test.py
boomanaiden154: It will error out as `registry.get_configuration` relies on being configured through gin and it doesn't get configured by pytest, so we need to mock it so that it doesn't rely on being configured through a gin file.

compiler_opt/rl/problem_configuration.py
boomanaiden154: I believe we should be able to. In the problem-specific implementation of this flag we can create a function that takes in a parameter of flags to return and then return that list from `flags_to_add` and `flags_to_delete` within that implementation. It's not a super elegant way to do it though. I can't really think of anything better, but that doesn't mean something better doesn't exist.

yundiqian, line 110: use Tuple here, same below

compiler_opt/tools/generate_default_trace_test.py
yundiqian: you can have gin configs following the code in line 112~115

i have the runner mock here because we don't want to really run compilation here; for you change, you don't need the mock

compiler_opt/tools/generate_default_trace_test.py
```diff
@@ -89,6 +90,13 @@ def test_api(self, mock_get_runner):
         output_performance_path=os.path.join(tmp_dir.full_path,
                                              'output_performance'),
     ):
+      with gin.unlock_config():
+        gin.parse_config_files_and_bindings(
+            config_files=['compiler_opt/rl/inlining/gin_configs/common.gin'],
+            bindings=[
```
yundiqian, line 96: bindings=None would work

compiler_opt/tools/generate_default_trace_test.py
```diff
@@ -89,6 +90,13 @@ def test_api(self, mock_get_runner):
         output_performance_path=os.path.join(tmp_dir.full_path,
                                              'output_performance'),
     ):
+      with gin.unlock_config():
```
yundiqian, line 93: does it need to happen here?

does it work if you move it to the setUp of the test class?

boomanaiden154: Moving it to the `setUp` function causes gin to complain that the config is locked. This makes sense as `unlock_config` creates a context manager designed to be used with a `with` statement, so all the context will get destroyed before the actual test functions get called. I don't see any easy way to create this context in a function that covers the entire test class.

yundiqian: have you tried that? gin complains here for config is locked as well, that's why there is the with gin.unlock_config() here

i think you only need the gin.parse_... in the gin.unlock_config() context for it to work. after running it, the gin overwritten is already happening, and you don't need the rest of your code to be in the context.

boomanaiden154: Moving both the `gin.unlock_config` statement and the config parsing to the `setUp()` function works. For some reason initially I only thought to move the `gin.unlock_config()` statement to the setup which definitely did not work. Thanks for the suggestion and your persistence.

PR #93, opened 2022-08-10
Prevent TF from logging info messages during data collection

boomanaiden154:
Currently, whenever a TF model gets loaded in by LLVM in the
development mode advisors, TF outputs a slew of info messages
to STDOUT that aren't particularly useful. This patch gets
rid of those messages by setting an environment variable when
running the compilation commands that only enables warnings
and errors to be printed to the console. This makes it much
easier to see progress/results in several of the scripts within
this repository such as train_locally.py and
generate_default_trace.py when grabbing performance data for a
policy.

### DIFF
--- a/compiler_opt/rl/compilation_runner.py
+++ b/compiler_opt/rl/compilation_runner.py
@@ -17,6 +17,7 @@
 import abc
 import dataclasses
 import json
+import os
 import subprocess
 import threading
 from typing import Dict, List, Optional, Tuple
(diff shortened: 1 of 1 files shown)


compiler_opt/rl/compilation_runner.py
```diff
@@ -152,8 +153,13 @@ def start_cancellable_process(
     TimeoutExpired: if the process times out.
     ProcessKilledError: if the process was killed via the cancellation token.
   """
+  command_env = os.environ.copy()
```
mtrofin, line 156: can you make this configurable via an abseil flag? I'd just do "quiet" (default) vs not quiet - i.e. no need to elaborate on other values of TF_CPP_MIN_LOG_LEVEL or anything.

mtrofin:
lgtm, just a nit. Thanks for doing this!

compiler_opt/rl/compilation_runner.py
boomanaiden154: Added a `--quiet`/`--noquiet` flag that toggles this.

PR #94, opened 2022-08-10
Add BC hyperparameter tuning script

boomanaiden154:
This PR adds in a script in the `compiler_opt/tools` directory called `bc_hyperparam_tuning.py` that adds the ability to tie into tensorboard's [hyperparameter tuning](https://www.tensorflow.org/tensorboard/hyperparameter_tuning_with_hparams) dashboard specifically for behavioral cloning (planning on implementing reinforcement training/maybe a combined pipeline soon).

### Design decisions/discussion
* There is some refactoring that I had to do (and probably need to do more of) in some of the other files so that I could tie into all the existing infrastructure (especially python files setup like scripts such as `generate_default_trace.py` and `train_bc.py`). I can implement whatever is decided upon as being most optimal (which I think would probably be refactoring some of the reused functions into separate files so that they can be reused without the hackiness of gating the flags with a `if __name__ == '__main__'` statement, but there are definitely other ways to do this. Current method was just to get something working.
* I'm currently tying into the default trace functionality to grab performance data about how the model performs to make sure it generalizes to the entire corpus (or a sufficient percentage of it). I added in grabbing the actual reward metric from the compilation runner and made the refactored `generate_default_trace` function return the arithmetic mean of these values since that is what tensorboard uses and the reward can be highly variable (<-1 or 1<)(at least in the regalloc case) which makes it impossible to use a geometric mean even though these are relative values and geomean should probably be used.
* To iterate through all the parameters, I'm currently just setting up `HParam` objects using the tensorboard APIs directly in the code and then iterating through all of the possible combinations with nested for loops. This isn't really highly configurable, but it is extremely easy to implement and given how often this needs to be run, I'm not sure making it configurable through a gin config or something is necessarily something super high priority and wouldn't even be that clean to implement. I'm open to ideas here though.


boomanaiden154:
@mtrofin @yundiqian Do you guys mind reviewing this when you get a chance? For whatever reason I can't request reviews through the proper channel on the Github UI.

compiler_opt/tools/bc_hyperparam_tuning.py
```diff
@@ -0,0 +1,92 @@
...
+    for num_units_layer2 in hp_num_units_layer2.domain.values:
+      hparams = {
+          'HP_NUM_UNITS_LAYER1': num_units_layer1,
+          'HP_NUM_UNITS_LAYER2': num_units_layer2
+      }
+      with gin.unlock_config():
+        gin.bind_parameter('RegAllocNetwork.fc_layer_params',
+                           (num_units_layer1, num_units_layer2))
+      output_path = os.path.join(_OUTPUT_DIR.value, f'./run{run_number}')
+      mean_reward = run_training()
```
yundiqian, line 77: if i understand it correctly, it only runs the next set of hyper parameters after the current one is finished? so one after another?

compiler_opt/tools/bc_hyperparam_tuning.py
```diff
@@ -0,0 +1,92 @@
...
+        root_dir=root_dir,
+        data_path=_DATA_PATH.value)
+    return generate_default_trace.generate_default_trace(
+        data_path=_CORPUS_PATH.value,
+        policy_path=os.path.join(root_dir, './saved_policy'),
+        sampling_rate=0.01)
+
+
+def run_tuning():
+  hp_num_units_layer1 = hp.HParam('num_units_layer1', hp.Discrete([40, 80]))
```
yundiqian, line 64: what's the extra functionality this hyper-parameter tuning can do? is it equivalent to a simple shell script?

compiler_opt/tools/bc_hyperparam_tuning.py
boomanaiden154: Yes. It runs them all sequentially. Running in them in parallel on the same machine probably wouldn't provide any speedup as the BC training is already heavily multithreaded. Running it in parallel distributed across multiple machines would help, but supporting that doesn't seem necessary currently.

compiler_opt/tools/bc_hyperparam_tuning.py
boomanaiden154: It does quite a bit more than what a shell script could. It ties in directly to tensorboard through the HParams API and it can be setup to do much more complex searching through the possible space of hyperparameters than brute force like it is set up to do currently (although that isn't implemented at all yet).

yundiqian: I checked the documentation here https://www.tensorflow.org/tensorboard/hyperparameter_tuning_with_hparams a bit. I don't see much extra functionality with this tool? e.g., smarter tuning algorithms.

the integration to tensorboard is not so significant actually

yundiqian: by looking at the code https://github.com/tensorflow/tensorboard/tree/78db135e980fff6852cd91f3da439b45e7f9fc1e/tensorboard/plugins/hparams

it seems that it is not something under active development, so i'm wondering whether it is the best tool to use?

boomanaiden154: That github link that you sent is to a commit that is pretty old. Looking at head-of-tree, it definitely seems like it isn't under active development, but it at least seems to be getting some maintenance (last commit in December of last year). 

In regards to your other comment, you're right in that the tensorboard integration specifically doesn't allow for better tuning algorithms. That would have to be implemented on our side, but it is an example of something that couldn't easily be done with a shell script. The integration isn't anything too significant, but I thought the visualization method that they use looked fairly useful for actually deciding what hyperparameters are important and what values are better. However, I'm not too familiar with any other options for tuning.

compiler_opt/tools/bc_hyperparam_tuning.py
```diff
@@ -0,0 +1,92 @@
...
+    'Gin bindings to override the values set in the config files.')
+
+METRIC_REWARD = 'reward'
+
+
+def run_training():
+  with tempfile.TemporaryDirectory() as temp_dir:
+    root_dir = os.path.expanduser(temp_dir)
+    root_dir = os.path.normpath(root_dir)
+    # TODO(boomanaiden154): leave num_iterations and batch_size to config
```
yundiqian, line 51: i guess they can be configured with gin file?

compiler_opt/tools/bc_hyperparam_tuning.py
```diff
@@ -0,0 +1,92 @@
...
+    train_bc.train_eval(
```
yundiqian, line 52: is it possible to run the train_locally script instead of calling the train_eval function? this will make things much easier

compiler_opt/tools/bc_hyperparam_tuning.py
```diff
@@ -0,0 +1,92 @@
...
+        data_path=_DATA_PATH.value)
+    return generate_default_trace.generate_default_trace(
+        data_path=_CORPUS_PATH.value,
+        policy_path=os.path.join(root_dir, './saved_policy'),
+        sampling_rate=0.01)
+
+
+def run_tuning():
+  hp_num_units_layer1 = hp.HParam('num_units_layer1', hp.Discrete([40, 80]))
+  hp_num_units_layer2 = hp.HParam('num_units_layer2', hp.Discrete([40, 80]))
```
yundiqian, line 65: is it possible to organize the hp setup in a more configurable way? the code here looks very adhoc

compiler_opt/tools/bc_hyperparam_tuning.py
yundiqian: i'm not familiar with open source smart tuning packages as well, you can try to search for "bayesian optimization hyperparameter tuning" to see whether there are any good open source ones

compiler_opt/tools/bc_hyperparam_tuning.py
```diff
@@ -0,0 +1,92 @@
...
+  with tempfile.TemporaryDirectory() as temp_dir:
+    root_dir = os.path.expanduser(temp_dir)
+    root_dir = os.path.normpath(root_dir)
+    # TODO(boomanaiden154): leave num_iterations and batch_size to config
+    train_bc.train_eval(
+        num_iterations=100,
+        batch_size=64,
+        root_dir=root_dir,
+        data_path=_DATA_PATH.value)
+    return generate_default_trace.generate_default_trace(
```
yundiqian, line 57: a better way to evaluate the behavior cloning policy is to look at the bc loss during training

compiler_opt/tools/bc_hyperparam_tuning.py
boomanaiden154: That should definitely be possible. I'll have to figure out exactly how to get it working (and the exact syntax we want), but it shouldn't be too difficult.

compiler_opt/tools/bc_hyperparam_tuning.py
boomanaiden154: Yes. Definitely pretty WIP in currently. The plan now is to make everything configurable in gin so through a gin config we can set the parameters we wish to tune and the potential values within the config and then just pass that to the script where it will either just iterate through them all or potentially do something more intelligent (still need to look into libraries/methods for that a little bit more).

compiler_opt/tools/bc_hyperparam_tuning.py
boomanaiden154: Should definitely be possible with the `subprocess` module. I'll work on changing over to that when I get a chance.

compiler_opt/tools/bc_hyperparam_tuning.py
boomanaiden154: Would you recommend taking the loss at the final iteration the model trains or doing something more fancy like taking the minimum loss value seen over all steps that get logged?  The loss values can be pretty jumpy towards the end from what I've seen (like changing by a factor of two between the second to last and last loss logs).

compiler_opt/tools/bc_hyperparam_tuning.py
boomanaiden154: Switched over to using `keras_tuner` which fits in well with a lot of our existing tooling. Using the bayesian optimizer available through that package.

compiler_opt/tools/bc_hyperparam_tuning.py
boomanaiden154: Switched to configuring the hyperparams to configure with a new gin config file, `compiler_opt/rl/reglloc/gin_configs/bc_hyperparam_tuning.gin`.

compiler_opt/tools/bc_hyperparam_tuning.py
boomanaiden154: Switched to calling everything with `subprocess` and parsing the data that we want from the tensorboard output logging.

compiler_opt/rl/train_bc.py
```diff
@@ -35,29 +35,30 @@
...
-    'data_path', [],
-    'Path to TFRecord file(s) containing training data. Skip training and dump'
-    'an untrained model with random weights (for testing purpose) if '
-    'unspecified.')
-_GIN_FILES = flags.DEFINE_multi_string(
-    'gin_files', [], 'List of paths to gin configuration files.')
-_GIN_BINDINGS = flags.DEFINE_multi_string(
-    'gin_bindings', [],
-    'Gin bindings to override the values set in the config files.')
+if __name__ == '__main__':
```
yundiqian, line 38: do you still need this and changing the train_eval interface after switching to subprocess?

compiler_opt/tools/bc_hyperparam_tuning.py
```diff
@@ -0,0 +1,150 @@
...
+  with tempfile.TemporaryDirectory() as temp_dir:
+    bc_script_command = [
+        'python3', 'compiler_opt/rl/train_bc.py', f'--root_dir={temp_dir}',
+        f'--data_path={_DATA_PATH.value}'
+    ]
+    bc_script_command = extend_command_with_gin_files(bc_script_command)
+    for hparam in hparams:
+      gin_name, hparam_value = hparam
+      bc_script_command.append(f'--gin_bindings={gin_name}={hparam_value}')
+    print(bc_script_command)
```
yundiqian, line 80: logging instead of print

compiler_opt/rl/trainer.py
```diff
@@ -86,6 +86,9 @@ def __init__(
     self._summary_writer.set_as_default()
 
     self._global_step = tf.compat.v1.train.get_or_create_global_step()
+    # reset the global step to zero so that it doesn't continually increase
+    # if we create multiple instances of trainer
+    self._global_step.assign(0)
```
yundiqian, line 91: no longer need it after changing to subprocess?

compiler_opt/rl/train_bc.py
boomanaiden154: Good catch. I thought I reverted all these changes previously, but apparently I didn't. Fixed now.

compiler_opt/rl/regalloc/gin_configs/network.gin
```diff
@@ -9,6 +9,8 @@ regalloc.config.get_observation_processing_layer_creator.with_z_score_normalizat
 create_agent.policy_network = @regalloc_network.RegAllocNetwork
 
 RegAllocNetwork.preprocessing_combiner=@tf.keras.layers.Concatenate()
-RegAllocNetwork.fc_layer_params=(80, 40)
+num_units_layer1=80
```
yundiqian, line 12: this change hurts the flexibility, suppose we want to run behavioral cloning of (40, 20) and (40, 20, 20), this is not going to work.

Is it possible to iterate between tuples? so [(40, 20), (40, 20, 20)]

boomanaiden154: It's not possible to directly iterate between tuples as the bayesian optimizer only wants to operate over single continuous (well, discrete for the integers, but I'm guessing it's floating inside the implementation) variables. It's definitely not particularly flexible, but since we can't add extra variables to the optimizer at runtime, the only other solutions that I can think of are pretty clunky and also somewhat inflexible (eg something like having a `num_layers` attribute and being able to set the maximum number of layers and just having extra variables for each per-layer parameter that are sometimes ignored depending upon the value of `num_layers`). Kind of clunky and the optimizer there would be working on variables that don't always have an impact. Shouldn't be a massive deal, but it will make the acquisition function less efficient.

yundiqian: does the bayesian optimizer supports some kind of category choice? we can treat it as categories instead of numerical

boomanaiden154: There is a category choice available (documentation [here](https://keras.io/api/keras_tuner/hyperparameters/#choice-method)), but tuples are not on the officially supported list of types that it will cycle through. I could implement as an integer/choice of an integer and then just map that integer to the choices made available in the gin config, but that's still a pretty clunky solution.

yundiqian: ic, in your config you can have [(80, 40), (40, 20, 20)] for it. For numerical stuff, you can have [2, 3].

In your code, you can check whether the content is an int or float or bool, if they are, keep them, or else convert them to string. The code does not look too clunky in this way i guess?

boomanaiden154: Yeah. Probably shouldn't be too bad. Converting everything to a string would probably make a decently clean implementation. I'll work on an implementation and see what I can come up with.

boomanaiden154: Just wrote up a preliminary implementation of this. I didn't end up using the choice option as if we want to have more than 1 per layer hyperparameter, it quickly becomes less feasible. The keras tuner package allows for dynamically grabbing hyperparameters per trial, so depending upon the trial it just asks for different parameters from the optimizer. Might be slightly less efficient for the optimizer, but it's probably better than writing our own stuff to concatenate all reasonable permutations of per layer hparams and piping that in as a choice for the model.

compiler_opt/rl/regalloc/gin_configs/network.gin
```diff
@@ -11,4 +11,4 @@ create_agent.policy_network = @regalloc_network.RegAllocNetwork
 RegAllocNetwork.preprocessing_combiner=@tf.keras.layers.Concatenate()
 RegAllocNetwork.fc_layer_params=(80, 40)
 RegAllocNetwork.dropout_layer_params=None
-RegAllocNetwork.activation_fn=@tf.keras.activations.relu
```
yundiqian, line 14: revert this change

yundiqian (approved):

boomanaiden154:
Just rebased and made the requested change. CI is currently failing as the typecheck job needs to have `keras-tuner` installed, but I currently have it disabled so it doesn't end up pulling in another tensorboard version that creates a conflict. We'll probably need to wait for #128 to go through to prevent breaking anything.

yundiqian:
> Just rebased and made the requested change. CI is currently failing as the typecheck job needs to have `keras-tuner` installed, but I currently have it disabled so it doesn't end up pulling in another tensorbo

sounds good!

boomanaiden154:
This is succeeded by 9cc035f9c3c950c75fddcba24a362f861caa0614. If we want a script to automatically do the hyperparameter search, we should set it up in xmanager.

PR #95, opened 2022-08-10
Utility library/script to convert saved models to tflite models.

mtrofin:
The core functionality is converting a saved model to tflite, but for
mlgo purposes, we also need the `output_spec.json` to be copied.


yundiqian (commented):
can you make it as a library instead of binary? i guess we will be calling these functions during training?

mtrofin:
> can you make it as a library instead of binary? i guess we will be calling these functions during training?

a, yes - it doesn't need to be a binary, really; I just have the "binary" scenario in llvm, where I have a copy of this anyway (for test)

compiler_opt/rl/saved_model_tools.py
```diff
@@ -0,0 +1,62 @@
...
+from tf_agents.policies import greedy_policy  # pylint: disable=unused-import
+
+
+def convert_saved_model(sm_dir: str, tflite_model_file_path: str):
+  """Convert a saved model to tflite.
+
+  Args:
+    sm_dir: path to the saved model to convert
+
+    tflite_model_file_path: desired output file path. Directory structure will
```
yundiqian, line 30: nit: tflite_model_path?

compiler_opt/rl/saved_model_tools.py
```diff
@@ -0,0 +1,62 @@
...
+
+    tflite_model_dir: path to a directory where the tflite model will be placed.
+    The model will be named model.tflite. Alongside it will be placed a copy of
+    the output_spec.json file.
+  """
+  tf.io.gfile.makedirs(tflite_model_dir)
+  convert_saved_model(mlgo_model_dir,
+                      os.path.join(tflite_model_dir, 'model.tflite'))
+
+  json_file = 'output_spec.json'
```
yundiqian, line 59: add the functions and test in policy_saver.py? 'output_spec.json' is defined as constant there.

compiler_opt/rl/saved_model_tools.py
```diff
@@ -0,0 +1,62 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Tools for manipulating saved models."""
+
+import tensorflow as tf
+import os
+
+# greedy_policy is needed during conversion
```
yundiqian, line 20: why this is needed during conversion? 

and we need to convert the collect_policy (so non greedy) during training

compiler_opt/rl/saved_model_tools.py
```diff
@@ -0,0 +1,62 @@
...
+    the saved model files (i.e. saved_model.pb, the variables dir) and the
+    output_spec.json file
+
+    tflite_model_dir: path to a directory where the tflite model will be placed.
+    The model will be named model.tflite. Alongside it will be placed a copy of
+    the output_spec.json file.
+  """
+  tf.io.gfile.makedirs(tflite_model_dir)
+  convert_saved_model(mlgo_model_dir,
+                      os.path.join(tflite_model_dir, 'model.tflite'))
```
yundiqian, line 57: define 'model.tflite' as constant?

compiler_opt/rl/saved_model_tools.py
mtrofin: done

compiler_opt/rl/saved_model_tools.py
mtrofin: done

compiler_opt/rl/saved_model_tools.py
mtrofin: Huh. Maybe it's an artifact of when I was incorrectly specifying how to carry out the conversion, when I was allowing undefined (TF-full defined) opcodes, and this was then required. Seems to work w/o. Removed.

compiler_opt/rl/saved_model_tools.py
mtrofin: Done

compiler_opt/rl/policy_saver.py
```diff
@@ -158,3 +159,44 @@ def save(self, root_dir: str):
...
+
+    tflite_model_dir: path to a directory where the tflite model will be placed.
+    The model will be named model.tflite. Alongside it will be placed a copy of
+    the output_spec.json file.
+  """
+  tf.io.gfile.makedirs(tflite_model_dir)
+  convert_saved_model(mlgo_model_dir,
+                      os.path.join(tflite_model_dir, TFLITE_MODEL_NAME))
+
+  json_file = 'output_spec.json'
```
yundiqian, line 199: use OUTPUT_SIGNATURE defined above

compiler_opt/rl/policy_saver.py
```diff
@@ -158,3 +159,44 @@ def save(self, root_dir: str):
...
+  converter = tf.lite.TFLiteConverter.from_saved_model(sm_dir)
+  converter.target_spec.supported_ops = [
+      tf.lite.OpsSet.TFLITE_BUILTINS,
+  ]
+  tfl_model = converter.convert()
+  with tf.io.gfile.GFile(tflite_model_path, 'wb') as f:
+    f.write(tfl_model)
+
+
+def convert_mlgo_model(mlgo_model_dir: str, tflite_model_dir: str):
```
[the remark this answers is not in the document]
mtrofin: we have that flexibility, but want to also have the ability to easily separate the 2 - otherwise you have to delete the saved model.

compiler_opt/rl/policy_saver.py
mtrofin: Ack - I'll patch subsequently, I basically forgot about this, deleted my branch, and now need this to actually get tflite models produced during training, since llvm supports that. So I'll land this, then patch accordingly.

PR #96, opened 2022-08-10
Introduce Validation Data Collector

Northbadge:
- Allows concurrent evaluation of models on a separate dataset during training, with `--validation_data_path`
- This is done with minimal impact on training time by only utilizing the CPU for the validation dataset when it is mostly idle doing tf.train(), and pinning processes to specific CPUs
- The amount of impact can be adjusted via a gin.config on cpu_affinity.py
- CPU affinities are only optimized for internal AMD-Zen based systems at the moment, but can be extended in the future.

Missing tests at the moment, will add after the interface is more concrete


Northbadge:
Stuff is gonna move around, don't worry about reviewing too in-depth for now

compiler_opt/distributed/local/cpu_affinity.py
```diff
@@ -0,0 +1,55 @@
...
+"""Utility functions to set cpu affinities when operating main and subprocesses
+simultaneously."""
+import gin
+import psutil
+import itertools
+
+_NR_CPUS = psutil.cpu_count()
+
+_CPU_CONFIG = {  # List of CPU numbers in cache-sharing order.
+    # 'google-epyc' assumes logical core 0 and N/2 are the same physical core.
```
mtrofin, line 24: it can be probably named something more neutral, like 'default'?

compiler_opt/rl/inlining/inlining_runner.py
```diff
@@ -71,6 +71,8 @@ def compile_fn(
       cancelled work.
       RuntimeError: if llvm-size produces unexpected output.
     """
+    if cancellation_manager is None:
```
mtrofin, line 74: weird... hmm, should we just not pass `cancellation_manager` because there's `self._cancellation_manager` anyway?

compiler_opt/rl/local_validation_data_collector.py
```diff
@@ -0,0 +1,243 @@
...
+  def __init__(self, cps: corpus.Corpus, worker_pool_args, reward_stat_map,
+               max_cpus):
+    self._num_modules = len(cps) if cps is not None else 0
+    self._corpus: corpus.Corpus = cps
+    self._default_rewards = {}
+
+    self._running_policy = None
+    self._default_futures: List[worker.WorkerFuture] = []
+    self._current_work: List[Tuple[corpus.ModuleSpec, worker.WorkerFuture]] = []
+    self._last_time = None
```
mtrofin, line 47: do we need the time stuff anymore?

Northbadge: the time stuff is purely cosmetic, just to get the total wall time spent compiling validation modules

compiler_opt/distributed/local/cpu_affinity.py
Northbadge: 'default' kind of implies it'd work fine for any system, whereas we can only guarantee it'll work fine on a google server running Epyc CPUs

Northbadge:
Rebase after #113 #114 #115 , might need some slight modification on _schedule_jobs depending on cancellation_manager.

PR #97, opened 2022-08-11
fix combine_training_corpus.py to use corpus_description.json

yundiqian:


compiler_opt/tools/combine_training_corpus.py
```diff
@@ -62,12 +63,16 @@ def main(argv):
       continue
 
     with tf.io.gfile.GFile(path, 'r') as f:
-      module_names.extend(
-          [os.path.join(sub_dir, name.rstrip('\n')) for name in f])
+      corpus_description = json.load(f)
+      module_names.extend([
```
mtrofin, line 67: merging corpora needs to be aware of whether they are different: today, one could have a global_command_override, and the other another or none; or, shortly, they could each need a different FDO profile (like afdo or CSFDO).

I think we should rather allow assembling corpora a'la what Jacob did last summer - i.e. allow our tools use an abstract assembling of corpora, which can be sampled, and upon sampling get the right command line, etc, and also whether anything needs pre-copied (like large profile files)

That may be more complicated, though, so if this unblocks something, let's at minimum add a validation that the 2 corpora are merge-able - i.e. they don't have any other fields than 'modules' would be a good start; also a mention of that in the docstring.

yundiqian: i agree that a better is to have the trainer be able to consume multiple corpus. However, this script would still be helpful in the case of e.g. combining spec corpus, where there are 20 same-configuration corpus to combine

compiler_opt/tools/combine_training_corpus.py
```diff
@@ -62,12 +67,20 @@ def main(argv):
...
+          os.path.join(sub_dir, name) for name in corpus_description['modules']
+      ])
+      del corpus_description['modules']
```
mtrofin, line 74: nit: is there some other way that's not destructive to do the check? like just check there's no other key than 'modules' - non-mutating code is easier to reuse (so if we end up trying to reuse this, one less surprise)

mtrofin:
lgtm, a nit

PR #98, opened 2022-08-11
Add raw_reward_only to runners

Northbadge:
- Will be useful for validation runner and generate default trace
- Allows None values in sequence_examples
- Adds a todo w.r.t. CompilationResult having odd asserts/redundant attributes

Part of #96


Northbadge:
reset the wrong branch to upstream main and it... auto recognized as merged...

PR #100, opened 2022-08-11
Add reward_only to runners

Northbadge:
- Will be useful for validation runner and generate default trace
- Adds a todo w.r.t. CompilationResult having odd asserts/redundant attributes

Part of #96 

(was #98 before I messed up my branch reset and GitHub took the opportunity to close it)

### DIFF
--- a/compiler_opt/rl/compilation_runner.py
+++ b/compiler_opt/rl/compilation_runner.py
@@ -287,8 +287,6 @@ class CompilationRunner(Worker):
 
   @staticmethod
   def get_rewards(result: Dict) -> List[float]:
-    if len(result) == 0:
-      return []
     return [v[1] for v in result.values()]
 
   def collect_results(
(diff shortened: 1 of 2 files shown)


mtrofin:
lgtm, but rename the patch because there's no "raw_" anywhere anymore

compiler_opt/rl/compilation_runner.py
```diff
@@ -275,6 +285,46 @@ def enable(self):
   def cancel_all_work(self):
     self._cancellation_manager.kill_all_processes()
 
+  @staticmethod
+  def get_rewards(result: Dict) -> List[float]:
+    return [v[1] for v in result.values()]
+
+  def collect_results(
```
yundiqian, line 292: i think it is over-engineering to refactor the collect_data function with the collect_results function, the options in the collect_results here are too complicated and the return type is also confusing, the name of collect_results and collect_data is also confusing.

I recommend to just write a function for your evaluation purpose, in this way the function args and return types will be much clearer

Northbadge: that's sounds reasonable. I'll just make `_compile_fn` public scoped, that's essentially all I need anyway

PR #101, opened 2022-08-11
Add pause/resume/context to workers

Northbadge:
- Allows a user to start/stop runners' subprocesses at will, via OS signals SIGSTOP and SIGCONT, in cancellation_manager.
- Allows a user to bind processes to specific CPUs, and modify their niceness.
- Allows local_worker_pool to be used outside of a context manager (by adding a destructor)
- Switch workers to be Protocol based, so Workers are effectively duck-typed (i.e. anything that has the required methods passes as a Worker)
- Fixes a memory leak in unregister_process

Part of #96

### DIFF
--- a/compiler_opt/distributed/local/local_worker_manager.py
+++ b/compiler_opt/distributed/local/local_worker_manager.py
@@ -31,8 +31,8 @@ import concurrent.futures
 import dataclasses
 import functools
 import multiprocessing
-import threading
 import psutil
+import threading
 
 from absl import logging
 # pylint: disable=unused-import
@@ -40,7 +40,7 @@ from compiler_opt.distributed.worker import Worker
 
 from contextlib import AbstractContextManager
 from multiprocessing import connection
-from typing import Any, Callable, Dict, Optional, List
+from typing import Any, Callable, Dict, List, Optional
 
 
 @dataclasses.dataclass(frozen=True)
--- a/compiler_opt/distributed/local/local_worker_manager.py
+++ b/compiler_opt/distributed/local/local_worker_manager.py
@@ -32,6 +32,9 @@ import dataclasses
 import functools
 import multiprocessing
 import threading
+import os
+import psutil
+import signal
 
 from absl import logging
 # pylint: disable=unused-import
@@ -39,7 +42,7 @@ from compiler_opt.distributed.worker import Worker
 
 from contextlib import AbstractContextManager
 from multiprocessing import connection
-from typing import Any, Callable, Dict, Optional
+from typing import Any, Callable, Dict, Optional, List
 
 
 @dataclasses.dataclass(frozen=True)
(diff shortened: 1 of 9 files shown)


Northbadge:
@petrhosek - I added a new dependency, psutil here

compiler_opt/distributed/local/local_worker_manager.py
```diff
@@ -205,10 +209,37 @@ def shutdown(self):
       try:
         # Killing the process triggers observer exit, which triggers msg_pump
         # exit
+        self.resume()
         self._process.kill()
       except:  # pylint: disable=bare-except
         pass
 
+    def pause(self):
```
mtrofin, line 217: I think you can have a test for pause/resume: the test worker keeps a list of timestamps. It adds every second to it. So you start it, wait 2 seconds, pause it, wait 5 seconds, resume, wait 2 seconds and read the list. There should be a >1 sec gap between 2 consecutive entries.

Northbadge:
so turns out this just sends SIGSTOP to the python worker and not the clang subprocess as well. I could either:
- have the clang subprocesses each be in their own pgrp, which makes .resume and .pause work on a per-worker basis
- or, have LocalValidationDataCollector spawned clang subprocesses share a single pgrp
But, either way, bash-sent signals would stop working since those get sent to the pgrp of the first process, and doing the above creates multiple pgrps

alternatively, could figure out a way to maintain a record of spawned clang processes PIDs, do the "cancel and requeue" version, or accept that some clang processes will run to completion when asked to stop. I'll probably get back to this next week.

compiler_opt/rl/compilation_runner.py
```diff
@@ -118,26 +129,70 @@ def register_process(self, p: 'subprocess.Popen[bytes]'):
     """Register a process for potential cancellation."""
     with self._lock:
       if not self._done:
-        self._processes.add(p)
+        self._processes[p.pid] = self.ProcData(
```
Northbadge, line 132: need to immediately SIGSTOP if self._paused here, will add later

compiler_opt/rl/compilation_runner.py
```diff
@@ -102,13 +104,22 @@ class WorkerCancellationManager:
   managing resources.
   """
 
-  def __init__(self):
+  @dataclasses.dataclass
+  class ProcData:
```
mtrofin, line 108: a thought: the motivating scenario here is the validator. For validation, we're actually OK to let compilation run longer than x seconds - because the goal is to get a thorough idea of what-if this model were shipped. So, how about:
- no ProcData
- just have the validator use a very large timeout, like 20 minutes (i.e. ~half of that in real compilation time)

compiler_opt/rl/compilation_runner.py
```diff
@@ -129,10 +131,33 @@ def kill_all_processes(self):
     for p in self._processes:
       kill_process_ignore_exceptions(p)
 
+  def pause_all_processes(self):
+    with self._lock:
+      if self._paused:
```
mtrofin, line 136: just saw this: you can just self._paused = True, that's the net outcome of this

PR #106, opened 2022-08-16
Add load balancer

### LINKED ISSUE #91: Tasks assigned to LocalWorkerPool are not load balanced
Tracking issue, search for this issue number in code.

- `_schedule_jobs` in `local_data_collector` load balances naively, in round-robin style
- Ideally, each worker is only assigned a couple jobs at a time, and gets more upon completion of previous tasks

Northbadge:
- Adds a buffered load balancer, which by default maintains at least 2 tasks assigned to each worker.

Closes #91

### DIFF
--- a/compiler_opt/distributed/local/buffered_scheduler.py
+++ b/compiler_opt/distributed/local/buffered_scheduler.py
@@ -74,7 +74,6 @@ def schedule(assignments: List[Callable[[T], concurrent.futures.Future]],
       # which is required to successfully completed the callback.
       results[i].add_done_callback(lambda _: threading.Thread(
           target=chain_assignments, args=(assignee,)).start())
-      # results[i].add_done_callback(lambda _: chain_assignments(assignee))
 
   for _ in range(buffer):
     for a in assignees:
--- a/compiler_opt/distributed/local/buffered_scheduler.py
+++ b/compiler_opt/distributed/local/buffered_scheduler.py
@@ -75,8 +75,8 @@ def schedule(work: List[Callable[[T], worker.WorkerFuture]],
 
       # This potentially causes a deadlock if chain_work is called via a
       # future.set_result() context which holds a resource that is also required
-      # to complete the call work[i](wkr) call above. See test_deadlock for an
-      # example.
+      # to complete the call work[i](wkr) call above. For an example, see:
+      # https://gist.github.com/Northbadge/a57f2d4e0a71e8f3934bdb47e59e343e
       # A fix/workaround would be using threading below, but that introduces
       # overhead of creating a new thread.
       results[i].add_done_callback(lambda _: chain_work(wkr))
--- a/compiler_opt/distributed/local/buffered_scheduler.py
+++ b/compiler_opt/distributed/local/buffered_scheduler.py
@@ -32,6 +32,8 @@ class DeadlockDetected(Exception):
 
 def raise_deadlock(*args, **kwargs):
   # This is raised if attempting to call .result() on _NEVER_FUTURE
+  # "Deadlock" here refers to blocking indefinitely, as the actual .result()
+  # would block indefinitely.
   raise DeadlockDetected()
 
 
@@ -60,7 +62,7 @@ def schedule(work: List[Callable[[T], worker.WorkerFuture]],
   idx_lock = threading.Lock()
 
   # Simple atomic increment and get.
-  # Used to iterate over work like a thread-safe queue.
+  # Used to iterate over `work` like a thread-safe queue without making a copy.
   def fetch_idx():
     nonlocal idx
     with idx_lock:
@@ -71,13 +73,13 @@ def schedule(work: List[Callable[[T], worker.WorkerFuture]],
     if (i := fetch_idx()) < len(work):
       results[i] = work[i](wkr)
 
-      # It is necessary for the done callback to execute on a separate thread,
-      # which is on an anonymous/fresh callstack. This is to prevent a potential
-      # deadlock arising from the current thread holding resources (e.g. a lock)
-      # which is required to successfully completed the callback.
-      # See: test_deadlock
-      results[i].add_done_callback(
-          lambda _: threading.Thread(target=chain_work, args=(wkr,)).start())
+      # This potentially causes a deadlock if chain_work is called via a
+      # future.set_result() context which holds a resource that is also required
+      # to complete the call work[i](wkr) call above. See test_deadlock for an
+      # example.
+      # A fix/workaround would be using threading below, but that introduces
+      # overhead of creating a new thread.
+      results[i].add_done_callback(lambda _: chain_work(wkr))
 
   for _ in range(buffer):
     for w in workers:
--- a/compiler_opt/distributed/local/buffered_scheduler.py
+++ b/compiler_opt/distributed/local/buffered_scheduler.py
@@ -19,7 +19,7 @@
 import concurrent.futures
 import threading
 
-from typing import List, Callable, TypeVar, Optional
+from typing import List, Callable, TypeVar
 
 from compiler_opt.distributed import worker
 
@@ -44,7 +44,6 @@ def schedule(work: List[Callable[[T], worker.WorkerFuture]],
   results = [concurrent.futures.Future() for _ in range(len(work))]
   idx = -1
   idx_lock = threading.Lock()
-  done = False
 
   # Simple atomic increment and get.
   # Used to iterate over `work` like a thread-safe queue without making a copy.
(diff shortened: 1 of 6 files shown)


compiler_opt/distributed/worker.py
```diff
@@ -28,26 +29,14 @@ def is_priority_method(cls, method_name: str) -> bool:
...
-class WorkerFuture(Generic[T], metaclass=abc.ABCMeta):
-
-  @abc.abstractmethod
-  def result(self) -> T:
-    raise NotImplementedError()
-
-  @abc.abstractmethod
-  def done(self) -> bool:
-    raise NotImplementedError()
+WorkerFuture = concurrent.futures.Future
```
Northbadge, line 32: This was needed to resolve pytype errors, since the scheduler was typed to accept any `concurrent.futures.Future`

compiler_opt/distributed/worker.py
```diff
@@ -28,26 +29,14 @@ def is_priority_method(cls, method_name: str) -> bool:
...
 
 
 def wait_for(futures: Iterable[WorkerFuture]):
   """Dask futures don't support more than result() and done()."""
   for f in futures:
-    try:
-      _ = f.result()
-    except:  # pylint: disable=bare-except
-      pass
+    while not f.done():
```
Northbadge, line 38: the scheduler returns a dummy future list which slowly gets replaced with real futures, so calling .result is not viable.

Northbadge (commented):
I went the route of returning a dummy future list first, mainly since we can't get the "true" future without calling a worker through its stub (which would assign said worker a job).
This comes with the limitation that the returned future list from the scheduler is "brittle" since it mustn't be modified (just needed some minor changes here and there to resolve), but preserves the expected outcome of `future.done()` and makes the future replacement implementation simple (it's literally 2 lines).

compiler_opt/distributed/worker.py
Northbadge: (and it turns out this doesn't work on 3.8, as I've discovered via github spamming emails, so don't merge this yet)

compiler_opt/distributed/local/buffered_scheduler.py
```diff
@@ -0,0 +1,82 @@
...
+_NEVER_FUTURE = concurrent.futures.Future()  # A Future that never completes.
+_NEVER_FUTURE.result = raise_deadlock
+T = TypeVar('T')
+
+
+def schedule(assignments: List[Callable[[T], concurrent.futures.Future]],
+             assignees: List[T],
+             buffer=2) -> List[concurrent.futures.Future]:
+  """
+  Assigns assignments to assignees once previous assignments of the assignee are
```
mtrofin, line 45: s/assignments/work

compiler_opt/distributed/worker.py
mtrofin: You may want to type it as WorkerFuture, though - see the comment in the 'before' code about Dask Futures.

compiler_opt/distributed/local/buffered_scheduler.py
```diff
@@ -0,0 +1,82 @@
...
+  # This is raised if attempting to call .result() on _NEVER_FUTURE
+  raise DeadlockDetected()
+
+
+_NEVER_FUTURE = concurrent.futures.Future()  # A Future that never completes.
+_NEVER_FUTURE.result = raise_deadlock
+T = TypeVar('T')
+
+
+def schedule(assignments: List[Callable[[T], concurrent.futures.Future]],
```
mtrofin, line 41: you may want this to be Callable[[T], WorkerFuture] - see the comment before about Dask's futures

compiler_opt/distributed/local/buffered_scheduler.py
mtrofin: also s/assignee/worker. Less terminology mapping.

compiler_opt/distributed/local/buffered_scheduler.py
```diff
@@ -0,0 +1,82 @@
...
+    A list of Futures that should be stored AS IS. Treat as a reference-to-list.
+  """
+  # Create "fake" futures to be returned first, these futures will be replaced
+  # with "real" futures later on.
+  results = [_NEVER_FUTURE] * len(assignments)
+  idx = -1
+  idx_lock = threading.Lock()
+
+  # Simple atomic increment and get.
+  def fetch_idx():
```
mtrofin, line 61: why not just use a queue that gets drained until empty()?

If the reason is "why allocate more", sgtm, and I'd then comment that `fetch_idx` + the list of work form a thread_safely dequeable queue, quicker to understand (i.e. the fact that fetch_idx does simple atomic increment is quite immediately obvious, but "why" is less so until one finishes reading the code)

compiler_opt/distributed/local/buffered_scheduler.py
```diff
@@ -0,0 +1,82 @@
...
+    nonlocal idx
+    with idx_lock:
+      idx += 1
+      return idx
+
+  def chain_assignments(assignee):
+    if (i := fetch_idx()) < len(assignments):
+      results[i] = assignments[i](assignee)
+
+      # It is necessary for the done callback to execute on a separate thread,
```
mtrofin, line 71: is this about `idx_lock`? How would it be held and not released, its release depends just on incrementing idx, and it's not waiting for anything else?

Northbadge: nope- see test_deadlock. 

The actual scenario was that the callback gets called upon future.set_result(), with that callstack- this happened to be in msg_pump of the local_worker stubs, which holds the lock to the future map while doing so. The callback function also wants to fetch the future map lock so it can place a new future on it, and there's the deadlock.

Technically, we can sacrifice a bit of correctness of the scheduler (by having it not spawn the thread and do the call directly) by releasing the lock in msg_pump before doing the set_result, but it'd be a pain to debug (from my experience) if it happens.

compiler_opt/distributed/local/buffered_scheduler_test.py
```diff
@@ -91,44 +91,44 @@ def task():
...
-    # Sleep for up to 1 second.
-    for _ in range(10):
-      if futures[0].done() and futures[1].done():
-        break
-      time.sleep(0.1)
-    # If >1s, and it's still not done, it's probably deadlocked.
-    self.assertTrue(futures[0].done())
-    self.assertTrue(futures[1].done())
-    t.join()
+  # def test_deadlock(self):
```
mtrofin, line 94: because we shouldn't check in commented code (as a matter of principle), how about peeling this in a gist and referencing that in the comment?

Also, could you please add a comment on LocalWorkerPool about the interaction w the scheduler - we can then move those comments to a common abstraction later, when that appears.

mtrofin: (lgtm otherwise)

compiler_opt/distributed/worker.py
```diff
@@ -30,24 +30,23 @@ def is_priority_method(cls, method_name: str) -> bool:
...
+      time.sleep(0.1)
```
mtrofin, line 49: could the dummy future be an actual concurrency.future.Future that resolves when the worker's future does?

compiler_opt/distributed/local/buffered_scheduler.py
```diff
@@ -0,0 +1,83 @@
...
+  def fetch_idx():
+    nonlocal idx
+    with idx_lock:
+      idx += 1
+      return idx
+
+  def chain_work(wkr: T,
+                 future: Optional[concurrent.futures.Future] = None,
+                 future_idx: int = 0):
+    nonlocal done
```
mtrofin, line 60: you probably don't need `done` - just let idx get bumped up

compiler_opt/distributed/local/buffered_scheduler.py
```diff
@@ -0,0 +1,83 @@
...
+      # A fix/workaround would be using threading below, but that introduces
+      # overhead of creating a new thread.
+      work[i](wkr).add_done_callback(lambda f: chain_work(wkr, f, i))
+    else:
+      done = True
+
+  # Use min() in case buffer is huge for some reason.
+  for _ in range(min(buffer, (len(work) // len(workers)) + 1)):
+    for w in workers:
+      chain_work(w)
```
mtrofin, line 81: would it be more readable if you did the stuff in lines 62-65 here, and then chain_work was always getting a `future` and `future_idx` value? 

can you also rename maybe `future` to `worker_future` and `future_idx` to 'result_future_idx` or smth like that; or for symmetry you can just pass the result future there instead of the index; or have this:

```
def make_handler(worker, result_future):
  def handler(worker_future):
    if (e:=worker_future.exception()) is not None:
       result_future.set_exception(e)
    else:
       result_future.set_result(worker_future.result())
  return hander
```
and then you `add_done_callback(make_handler(wkr, results[i]))`

Northbadge: got it, neat solution!

(It's a bit shocking how close to javascript we're writing Python)

mtrofin: you take that back! :)

compiler_opt/distributed/worker.py
```diff
@@ -30,16 +29,17 @@ def is_priority_method(cls, method_name: str) -> bool:
...
 
-  @abc.abstractmethod
   def result(self) -> T:
     raise NotImplementedError()
 
-  @abc.abstractmethod
   def done(self) -> bool:
     raise NotImplementedError()
 
+  def add_done_callback(self, fn) -> None:
```
mtrofin, line 40: ugh... dask. OK, let's move forward with the change, though, we'll sort out Dask after (like we can have that configuration pay with a handler thread, for example)

PR #111, opened 2022-08-18
Add doc for thinlto corpus extraction

Northbadge:


yundiqian (approved):

docs/thinlto.md
```diff
@@ -0,0 +1,42 @@
+# Extracting Corpora for ThinLTO-enabled Projects
+
+These steps are tailored for the inlining problem.
```
mtrofin, line 3: They would work otherwise, right? (with the right adjustments). Maybe just say that these steps use inlining -Oz  as example. 

...but are they inlining-specific?

mtrofin (approved):

PR #113, opened 2022-08-19
Modify Corpus to store module specs as a tuple

Northbadge:
Part of #96

### DIFF
--- a/compiler_opt/rl/corpus.py
+++ b/compiler_opt/rl/corpus.py
@@ -122,6 +122,10 @@ class Corpus:
     """Filters module specs, keeping those which match the provided pattern."""
     self._module_specs = [ms for ms in self._module_specs if p.match(ms.name)]
 
+  @property
+  def modules(self):
+    return list(self._module_specs)
+
   def __len__(self):
     return len(self._module_specs)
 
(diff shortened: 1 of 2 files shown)


compiler_opt/rl/corpus.py
```diff
@@ -122,6 +122,10 @@ def filter(self, p: re.Pattern):
     """Filters module specs, keeping those which match the provided pattern."""
     self._module_specs = [ms for ms in self._module_specs if p.match(ms.name)]
 
+  @property
+  def modules(self):
+    return list(self._module_specs)
```
mtrofin, line 127: why constrain as list?

Northbadge: you mean the list()? So a copy is returned, to prevent mutation of the internal list

mtrofin: a, ok. so `get_modules_copy`? and it's not a @property then?

compiler_opt/rl/corpus.py
```diff
@@ -86,20 +86,23 @@ def __init__(self,
                data_path: str,
                additional_flags: Tuple[str, ...] = (),
                delete_flags: Tuple[str, ...] = ()):
-    self._module_specs = _build_modulespecs_from_datapath(
-        data_path=data_path,
-        additional_flags=additional_flags,
-        delete_flags=delete_flags)
+    self.module_specs = tuple(
```
yundiqian, line 89: make it private and use @property to expose it instead?

yundiqian (approved):

PR #115, opened 2022-08-19
Default to using self.cancellation_manager

Northbadge:
Part of #96


compiler_opt/rl/inlining/inlining_runner.py
```diff
@@ -71,6 +71,9 @@ def compile_fn(
       cancelled work.
       RuntimeError: if llvm-size produces unexpected output.
     """
+    if cancellation_manager is None:
```
mtrofin, line 74: why even pass it as an arg, why not we remove it from the arglist and use self._cancellation_manager?

Northbadge: changed it, prob should check with yundi@ to see if that's an ok thing to do

PR #117, opened 2022-08-23
Add chromium patch

boomanaiden154:
This PR adds the chromium patch that was worked on in [this CL](https://chromium-review.googlesource.com/c/chromium/src/+/3731216) that we have decided to maintain downstream. I've put it in the `./experimenal` directory currently. Let me know if it would be better off somewhere else.

### DIFF
--- a/experimental/chromium-bitcode-embedding.patch
+++ b/experimental/chromium-bitcode-embedding.patch
@@ -1,0 +1,93 @@
+diff --git a/build/config/clang/clang.gni b/build/config/clang/clang.gni
+index 1e662184872b3..4aed699d271b4 100644
+--- a/build/config/clang/clang.gni
++++ b/build/config/clang/clang.gni
+@@ -15,4 +15,7 @@ declare_args() {
+       default_toolchain != "//build/toolchain/cros:target"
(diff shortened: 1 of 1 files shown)


experimental/chromium-bitcode-embedding.patch
```diff
@@ -0,0 +1,105 @@
+diff --git a/AUTHORS b/AUTHORS
+index 4b62a980e90c9..ac00eb5c5c859 100644
+--- a/AUTHORS
```
mtrofin, line 3: do we need the change to AUTHORS - since this is likely used a s an on the fly patch, at most it may cause patching problems

boomanaiden154: Good point. Definitely something unnecessary that could potentially hurt maintainability. I've removed it from the patch.

boomanaiden154:
Just adjusted the patch to not change a lot of the compilation settings that I was adjusting before as they all work fine when `-Xclang -fembed-bitcode=all` is used, but not when `-fembed-bitcode=all` is used. We want the former, so might as well make the patch simpler.

PR #118, opened 2022-08-23
Prepare requirements for tflite

mtrofin:
We need to get past TF 2.9, but 2.10 isn't released yet, and there are
compat issues between 2.10-RC1 and tfagents. The recommended approach is
to install nightlies of these.

### DIFF
--- a/compiler_opt/rl/constant_value_network.py
+++ b/compiler_opt/rl/constant_value_network.py
@@ -45,9 +45,8 @@ class ConstantValueNetwork(network.Network):
 
     self._constant_output_val = constant_output_val
 
-  def call(self, observation, step_type=None, network_state=(), training=False):
+  def call(self, inputs, step_type=None, network_state=(), training=False):
     _ = (step_type, training)
-    shape = nest_utils.get_outer_array_shape(observation,
-                                             self._input_tensor_spec)
+    shape = nest_utils.get_outer_array_shape(inputs, self._input_tensor_spec)
     return tf.constant(
         self._constant_output_val, tf.float32, shape=shape), network_state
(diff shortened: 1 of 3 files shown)


requirements.txt
```diff
@@ -20,7 +20,7 @@ h5py==3.6.0
 idna==3.2
 keras==2.7.0
```
ebrevdo, line 21: i think you can remove these?  keras*

ebrevdo: that's my guess.  let's retrigger tests and see?

mtrofin: huh. now they fail with the `StructuredValue` failure. Not _quite_ the same as https://github.com/google/ml-compiler-opt/runs/7932290113?check_suite_focus=true, though.

compiler_opt/rl/constant_value_network.py
```diff
@@ -45,9 +45,8 @@ def __init__(self, input_tensor_spec, constant_output_val=0, name=None):
 
     self._constant_output_val = constant_output_val
 
-  def call(self, observation, step_type=None, network_state=(), training=False):
+  def call(self, inputs, step_type=None, network_state=(), training=False):
```
mtrofin, line 48: @yundiqian this takes care of 

```
compiler_opt/rl/constant_value_network.py:48:2: W0237: Parameter 'inputs' has been renamed to 'observation' in overridden 'ConstantValueNetwork.call' method (arguments-renamed)
```

Anything I'm missing?

ebrevdo:
LGTM

ebrevdo (approved):

compiler_opt/rl/constant_value_network.py
yundiqian: the change looks good to me, original warning due to some trivial tf agents hassles

yundiqian (approved):

PR #120, opened 2022-08-25
TFLite: when saving a saved model, also save the tflite model

mtrofin:
Currently just saving it next to the saved model, so we can start
iterating using tflite models with llvm. Eventually we can get more
efficient.

### DIFF
--- a/compiler_opt/rl/policy_saver.py
+++ b/compiler_opt/rl/policy_saver.py
@@ -69,6 +69,46 @@ def _get_non_identity_op(tensor):
   return tensor
 
 
+def convert_saved_model(sm_dir: str, tflite_model_path: str):
+  """Convert a saved model to tflite.
+
+  Args:
+    sm_dir: path to the saved model to convert
+
+    tflite_model_path: desired output file path. Directory structure will
+    be created by this function, as needed.
+  """
+  tf.io.gfile.makedirs(os.path.dirname(tflite_model_path))
+  converter = tf.lite.TFLiteConverter.from_saved_model(sm_dir)
+  converter.target_spec.supported_ops = [
+      tf.lite.OpsSet.TFLITE_BUILTINS,
+  ]
+  tfl_model = converter.convert()
+  with tf.io.gfile.GFile(tflite_model_path, 'wb') as f:
+    f.write(tfl_model)
+
+
+def convert_mlgo_model(mlgo_model_dir: str, tflite_model_dir: str):
+  """Convert a mlgo saved model to mlgo tflite.
+
+  Args:
+    mlgo_model_dir: path to the mlgo saved model dir. It is expected to contain
+    the saved model files (i.e. saved_model.pb, the variables dir) and the
+    output_spec.json file
+
+    tflite_model_dir: path to a directory where the tflite model will be placed.
+    The model will be named model.tflite. Alongside it will be placed a copy of
+    the output_spec.json file.
+  """
+  tf.io.gfile.makedirs(tflite_model_dir)
+  convert_saved_model(mlgo_model_dir,
+                      os.path.join(tflite_model_dir, TFLITE_MODEL_NAME))
+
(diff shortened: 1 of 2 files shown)


compiler_opt/rl/policy_saver.py
```diff
@@ -69,6 +69,46 @@ def _get_non_identity_op(tensor):
   return tensor
 
 
+def convert_saved_model(sm_dir: str, tflite_model_path: str):
```
mtrofin, line 72: these need to be before the use.

compiler_opt/rl/policy_saver.py
```diff
@@ -69,6 +69,46 @@ def _get_non_identity_op(tensor):
...
+
+    tflite_model_dir: path to a directory where the tflite model will be placed.
+    The model will be named model.tflite. Alongside it will be placed a copy of
+    the output_spec.json file.
+  """
+  tf.io.gfile.makedirs(tflite_model_dir)
+  convert_saved_model(mlgo_model_dir,
+                      os.path.join(tflite_model_dir, TFLITE_MODEL_NAME))
+
+  src_json = os.path.join(mlgo_model_dir, OUTPUT_SIGNATURE)
```
mtrofin, line 107: addressed your comment from the previous PR which I needed to merge because I managed to delete my fork

compiler_opt/rl/policy_saver.py
```diff
@@ -69,6 +69,47 @@ def _get_non_identity_op(tensor):
...
+  with tf.io.gfile.GFile(tflite_model_path, 'wb') as f:
+    f.write(tfl_model)
+
+
+def convert_mlgo_model(mlgo_model_dir: str, tflite_model_dir: str):
+  """Convert a mlgo saved model to mlgo tflite.
+
+  Args:
+    mlgo_model_dir: path to the mlgo saved model dir. It is expected to contain
+    the saved model files (i.e. saved_model.pb, the variables dir) and the
```
ebrevdo, line 97: nit: two-space indent for this and the line below, also L101, L102

compiler_opt/rl/policy_saver.py
```diff
@@ -69,6 +69,47 @@ def _get_non_identity_op(tensor):
...
 
 
+def convert_saved_model(sm_dir: str, tflite_model_path: str):
+  """Convert a saved model to tflite.
+
+  Args:
+    sm_dir: path to the saved model to convert
+
+    tflite_model_path: desired output file path. Directory structure will
+    be created by this function, as needed.
```
ebrevdo, line 79: nit: 2 space indent this line.

ebrevdo (approved):

compiler_opt/rl/policy_saver.py
```diff
@@ -69,6 +69,47 @@ def _get_non_identity_op(tensor):
...
+  converter.target_spec.supported_ops = [
+      tf.lite.OpsSet.TFLITE_BUILTINS,
+  ]
+  converter.allow_custom_ops = True
+  tfl_model = converter.convert()
+  with tf.io.gfile.GFile(tflite_model_path, 'wb') as f:
+    f.write(tfl_model)
+
+
+def convert_mlgo_model(mlgo_model_dir: str, tflite_model_dir: str):
```
yundiqian, line 92: where is this function used? somewhere else?

mtrofin: just test, for now. we may need to refactor things a bit, but that would grow the scope of this patch.

compiler_opt/rl/policy_saver.py
mtrofin: done

compiler_opt/rl/policy_saver.py
mtrofin: done

yundiqian (approved):

PR #123, opened 2022-08-26
Migrate to Python 3.10

boomanaiden154:
This PR changes the Github CI to test against Python 3.9 and 3.10 instead of 3.8 and 3.9. Could be adjusted if we still want to support 3.8 though. Also adjusts the versions of some dependencies so that they are compatible with 3.10. There are a couple new deprecation warnings internal to tensorflow that I've excluded in the `pytest.ini`. Nothing super major and I'm guessing it will get fixed by the time anything gets even close to being deprecated.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -20,10 +20,10 @@ jobs:
         run: |
           if [ -z $ACT ]
           then 
-            _ver="[3.8, 3.9]"
+            _ver="['3.8', '3.9', '3.10']"
             _cache="1"
           else 
-            _ver="[3.8]"
+            _ver="['3.8']"
             _cache="0"
           fi
           echo "::set-output name=version_matrix::$_ver"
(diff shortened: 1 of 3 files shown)


mtrofin:
We want 3.8 because Fuchsia and Chrome use that. Fine to expand to 3.8, 3.9, and 3.10.

.github/workflows/main.yml
```diff
@@ -20,10 +20,10 @@ jobs:
         run: |
           if [ -z $ACT ]
           then 
-            _ver="[3.8, 3.9]"
+            _ver="['3.9', '3.10']"
```
mtrofin, line 23: keep 3.8

.github/workflows/main.yml
```diff
@@ -20,10 +20,10 @@ jobs:
...
             _cache="1"
           else 
-            _ver="[3.8]"
+            _ver="['3.10']"
```
mtrofin, line 26: leave to 3.8 - it's where the most errors will likely pop, stuff like type annotations. 3.10 is what probably people have already, so they can just run pytest and whatnot by hand.

PR #124, opened 2022-08-26
Update dockerfiles

boomanaiden154:
Update the dockerfiles to take advantage of the fact that we can now use newer versions of Python. The base image gets bumped to Ubuntu 22.04 which ships Python 3.10 by default, and the development docker image has all of the TFLite stuff ready to compile LLVM with in the `/tflite` directory.

Currently blocked on #123 due to some of the versions that need to be adjusted so that everything installs smoothly when using Python 3.10.


boomanaiden154:
Unblocked now as #123 just got merged.

PR #126, opened 2022-08-29
Add regalloc demo

boomanaiden154:
This PR adds a demo similar to the one for the inlining case for the ML regalloc eviction model based on a Chromium corpus. Dependent upon #117 being merged, and it currently still needs validation (in terms of making sure all of the commands are 100% accurate).


boomanaiden154:
* [ ] Add in information on replacing PGO flags.

docs/regalloc-demo/demo.md
```diff
@@ -0,0 +1,363 @@
...
+
+```bash
+cd /ml-compiler-opt
+```
+
+To start off training, we need to extract a per-object bitcode/compilation
+flag corpus. This can be done with the following script:
+
+```bash
+PYTHONPATH=$PYTHONPATH:. python3 compiler_opt/tools/extract_ir.py \
```
yundiqian, line 221: is chrome of thinlto build?

docs/regalloc-demo/demo.md
```diff
@@ -0,0 +1,363 @@
...
+  --cmd_filter="^-O2|-O3" \
+  --input=/chromium/src/out/Release/compile_commands.json \
+  --input_type=json \
+  --llvm_objcopy_path=/llvm-build/bin/llvm-objcopy \
+  --output_dir=corpus
```
yundiqian, line 226: do you want to make it a variable as the fuchsia demo?

docs/regalloc-demo/demo.md
```diff
@@ -0,0 +1,363 @@
...
+
+### Collect the Default Trace and Generate Vocab
+
+Before we run reinforcement training, it is best to train the model using
+behavioral cloning on the default heuristic. First off, we need to collect
+a trace of the decisions the default heuristic is making:
+
+```bash
+PYTHONPATH=$PYTHONPATH:. python3 compiler_opt/tools/generate_default_trace.py \
+  --data_path=/corpus \
```
yundiqian, line 237: same, variables for data_path and output_path?

docs/regalloc-demo/demo.md
```diff
@@ -0,0 +1,363 @@
...
+but it definitely doesn't hurt to regenerate them. If adding/modifying
+features, it is necessary to regenerate the vocab.
+
+Now that the vocab is present (or has been regenerated) and we have a default
+trace, we can start to train the model using behavioral cloning to mimic the
+default heuristic:
+
+```bash
+PYTHONPATH=$PYTHONPATH:. python3 compiler_opt/rl/train_bc.py \
+  --root_dir=/warmstart \
```
yundiqian, line 270: same here for the root_dir here, probably define a working dir as variables so the user can define it whatever they like, then have hardcoded sub-folder under the root working dir, e.g., $WORKING_DIR/warmstart, $WORKING_DIR/trace, etc

docs/regalloc-demo/demo.md
boomanaiden154: I'm currently not doing a ThinLTO build just because the patch that I have prepared currently is designed to work without ThinLTO. I need to write up a patch to add the necessary flags to extract a corpus using ThinLTO to do some validation for the Chromium patch, so I'll try and add a section on ThinLTO corpus extraction once that is ready.

docs/regalloc-demo/demo.md
boomanaiden154: I just modified the demo to use `$WORKING_DIR` environment variable and then based everything off of that. Probably makes things cleaner than having separate variables for everything. Thanks for the suggestion.

docs/regalloc-demo/demo.md
boomanaiden154: Just added info on thinLTO corpus extraction.

boomanaiden154:
I added sections/modified the demo to demonstrate thinLTO corpus extraction in addition to corpus extraction in the non-thinLTO case. Just tested it out and everything seems to be working. Only thing that seems to be kind of concerning is the percentage of modules that `extract_ir.py` is able to probably access. It seems to only be getting about half of the modules due to a lot of files not being in the locations that it expects (ie files included in a thin archive get named differently) and some third party code getting linked in and not getting built with thinLTO. Need to do some more investigation, but for now everything seems to be working decently well.

boomanaiden154:
Now that corpus extraction is working properly for the thinLTO case after changing some flags, there shouldn't be any more issues with this PR. Changing status away from WIP.

yundiqian (approved):

PR #127, opened 2022-09-01
Fix trainer only logging tensorboard data on summary_export_interval

boomanaiden154:
In #62 I separated out the summary_interval variable to
summary_log_interval and summary_export_interval, assuming that
tensorboard only flushed everything to disk when the callable passed to
tf.summary.record_if was true while still allowing writing to the
buffers. However, this assumption turned out to be mistaken as anytime
summary_export_interval and summary_log_interval weren't both hit by the
current step, nothing would be logged, so data would only get logged
when summary_export_interval was hit. This patch fixes that by getting
rid of summary_export_interval and changing everything back to
summary_log_interval while gating the actual writing of the scalars
using tf.summary.should_record_summaries() to keep the performance gains
mentioned in #62.

### DIFF
--- a/compiler_opt/rl/trainer.py
+++ b/compiler_opt/rl/trainer.py
@@ -55,7 +55,6 @@ class Trainer(object):
       checkpoint_interval=10000,
       log_interval=100,
       summary_log_interval=100,
-      summary_export_interval=1000,
       summaries_flush_secs=10):
     """Initialize the Trainer object.
 
@@ -69,8 +68,6 @@ class Trainer(object):
       log_interval: int, the training step interval for logging.
       summary_log_interval: the number of steps in between logging metrics
         to tensorboard.
-      summary_export_interval: int, the training step interval for exporting
-        to tensorboard.
       summaries_flush_secs: int, the seconds for flushing to tensorboard.
     """
     self._root_dir = root_dir
@@ -79,7 +76,6 @@ class Trainer(object):
     self._checkpoint_interval = checkpoint_interval
     self._log_interval = log_interval
     self._summary_log_interval = summary_log_interval
-    self._summary_export_interval = summary_export_interval
 
     self._summary_writer = tf.summary.create_file_writer(
         self._root_dir, flush_millis=summaries_flush_secs * 1000)
@@ -123,7 +119,7 @@ class Trainer(object):
 
   def _update_metrics(self, experience, monitor_dict):
     """Updates metrics and exports to Tensorboard."""
-    if tf.math.equal(self._global_step % self._summary_log_interval, 0):
+    if tf.summary.should_record_summaries():
       is_action = ~experience.is_boundary()
 
       self._data_action_mean.update_state(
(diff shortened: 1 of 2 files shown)


compiler_opt/rl/trainer.py
```diff
@@ -123,7 +119,7 @@ def _initialize_metrics(self):
 
   def _update_metrics(self, experience, monitor_dict):
     """Updates metrics and exports to Tensorboard."""
-    if tf.math.equal(self._global_step % self._summary_log_interval, 0):
+    if tf.summary.should_record_summaries():
```
yundiqian, line 122: I think what the current code is doing is that the number (self._data_action_mean) gets updated every self._summary_log_interval (e.g., 100), but gets writen to tensorboard every self._summary_export_interval (e.g., 1000). In this way the mean values are gets better tracked in a frequently that is more frequent than exporting.

With your change, the number update stays the same and it exports to tensorboard more frequently (1000->100). This may increase the training time a little bit (I don't timed so I'm not sure how large is the impact)

what is the reason you want more frequent logging to tensorboard?

boomanaiden154: Interesting that it might be taking the average of those values. I'll have to investigate that further. When I benchmarked it, the training time was extremely similar (both of them were doing about 160 steps/second plus or mins a couple on my machine).

The main motivation behind this patch was for #94, as if we're averaging the last x% of loss values, only having one loss value per 1000 steps (10% of the overall training run) seems a little high. If everything is getting averaged though, it should probably be fine to leave the way it is (although maybe with some slight indexing changes as the last loss value at 10k steps should get exported, but it isn't getting exported.

When I wrote the original patch, my intention was also to log data every 100 steps (`summary_log_interval`) and have that show up on tensorboard with that resolution while leaving the actually flushing of buffers to the file system to every 1000 steps (`summary_export_interval`), but I must've missed what exactly the `tf.summary` functions that we're calling do.

yundiqian: It looks like flushing to tensorboard follows summaries_flush_secs. Probably that's why the speed stays the same for you.

Exporting to tensorboard every 100 steps 10x the data size, which we don't really need in most cases. How about just passing in --gin_bindings=Trainer.summary_export_interval=100 in your bc tuning script?

boomanaiden154: I can do that. Since flushing to tensorboard is bound by `summaries_flush_secs` rather than `summary_export_interval`, I'm assuming we should use this patch since `summary_export_interval` and `summary_log_interval` are essentially two different gates on the same action, just changing `summary_log_interval to have a default value of `1000`?

yundiqian: my understanding is that summary_log_interval controls the update frequency of the tf.keras.metrics.mean(), summary_export_interval controls the density of data points (dots) in tensorboard. It makes sense that summary_log_interval is a number smaller so that the mean metrics are more frequently updated.

so summary_export_interval and summaries_flush_secs are different, every time the flushing is triggered, it exports a bunch of information to the file (then shows on tensorboard), the larger summary_export_interval, the smaller amount of data it is

boomanaiden154: Looking at the code again, this makes significantly more sense. Not sure why I couldn't see that before. Thank you.

PR #128, opened 2022-09-02
Move to using lockfiles and pip install --no-deps

boomanaiden154:
This patch moves from using a somewhat standard `requirements.txt` (it seems to have been a lockfile at somepoint in the past) to making all of the `requirements*.txt` files into lockfiles and then installing with `pip3 install --no-deps`. This is done for two reasons:

* It gives some more flexibility in what dependencies of what packages we're installing. This is essentially in #94 so that we don't end up with two separate tensorboard installations (`tensorboard` and `tb-nightly`) due to how upstream has setup their dependencies.
* It allows for more reproducible builds as all packages are tagged with a specific version.

This PR also includes a couple other minor improvements:
* upgrade `grpcio` - With the version that we were using previously, there was no downloadable version for Python 3.10, meaning it got compiled from source everytime the CI ran. This made the Python 3.10 jobs take significantly longer to run.
* Removed quiet switch from pip - This makes debugging easier and doesn't really hurt anything given that the output of each step in Github actions is collapsed by default and the pip installation is in its own step.
* Added a new warning to pytest ignore from tf_agents. Some package version changed and it popped up (also somewhat hard to debug given it only shows up on Python 3.10.6 and not Python 3.10.4).


boomanaiden154:
Everything should already be there, but after the transition to TFLite where we bumped some of the package versions (particularly the tensorflow stuff), some other versions got out of sync, so I've regenerated the lock file using `pip freeze`, and we never did it in the first place for the `requirements-dev.txt` file. 

I'm assuming by chunking you mean splitting this over multiple PRs?

I'm not totally sure why exactly the new warning popped up. Something with how this specific set of package versions match up. I'll do more investigation.

boomanaiden154:
Finally got all the dependencies playing nice with each other here. I've chunked them within the requirements files specifically to differentiate between what packages we're explicitly requiring and what packages are version-locked dependencies of those packages. Hopefully I've chunked them the way you've expected, but if not I can move stuff around. I'm still not completely sure why the new warning appears. I can't replicate it locally at all, and it only shows up in the CI. It's another distutils deprecation warning internal to tf-agents, so it shouldn't impact anything, but something to take note of.

I'm thinking it would also be good to have some documentation on updating dependencies so that it's easier whenever we have to do this again?

mtrofin:
@petrhosek had some ideas about lockfiles, adding him here.

boomanaiden154:
@petrhosek gentle bump.

petrhosek:
@boomanaiden154 see #175.

boomanaiden154:
Superseded by #175.

PR #130, opened 2022-09-03
Misc improvements to benchmarking infrastructure

boomanaiden154:
This PR makes a couple minor improvements for the benchmarking code.
  - Actually checks that the tests being run pass.
  - Added more tests from the Chromium unittests and removed the
    chromium browser_tests that won't run headlessly.
  - Add a new utility, filter_tests.py, that runs through all of the
    tests in a test suite description and makes sure they pass.
  - Fixed list_gtests.py (single character fix, main didn't have a
    parameter before)
  - Add some info on the new utility to the documentation
  - Add debugging info in benchmark_report_converter for when there is
    a difference between two benchmark sets, it actually prints out
    what the exact difference is.
  - Remove advisor flag from list of required flags when building the
    LLVM test suite as it isn't required when the no compile flags are
    set.


boomanaiden154:
Changing this to a work in progress. There seems to be some issues with the unit test suites not showing any good signals, perhaps due to a high amount of overhead relative to the other tests that are included in the Chromium perf tests. I'm planning on doing some more investigation and acting on those findings.

boomanaiden154:
After some more analysis, all of the unittests do produce good signals with the register allocator and are even more stable than some of the other test suites, along with covering more code paths, so we can probably keep them in. My earlier results mostly seem to be stemming from bad usage of the report conversion script.

mtrofin:
@boomanaiden154 just checking, this good to go?

boomanaiden154:
Yes. Everything in this patch has been validated when I was testing out using the ML register allocator on Chromium. I was seeing some weird behavior in the `benchmark_report_converter.py` script in regards to the actual values it would output, but that's a completely separate issue and might have just been due to how I was processing the resulting data.

Issue #129, opened 2022-09-03
Convert documentation to TFLite


boomanaiden154:
Some of the documentation that is around currently, particularly the inliner demo and some of the benchmarking documentation, as well as some of the benchmarking code itself, are stuck with the Tensorflow C API rather than TFLite. This documentation/code (just benchmarking) should be updated to work with TFLite.

I should be able to do this, just trying to find some time.

* [x] Update inlining demo
* [x] Update benchmarking code/documentation to work with TFLite

PR #131, opened 2022-09-04
Update Oz inlining demo to work with TFLite

boomanaiden154:
Some minor changes to the inlining demo, just changing all of the info
on using the Tensorflow C library to using TFLite, including adding info
on compilation of the library using the buildbot script.


boomanaiden154:
Should complete one task in #129.

petrhosek:
The [`fuchsia.xml` manifest](https://github.com/google/ml-compiler-opt/blob/main/docs/demo/fuchsia.xml) we have in the repository isn't compatible with TFLite because it pins an older revision of LLVM toolchain that doesn't contain the TFLite support. We need to either remove it or change the instructions to cover both cases, that is pinned Fuchsia manifest and Fuchsia at `HEAD`. I'm leaning slightly towards the former.

boomanaiden154:
I'm learning towards the former too, especially since quite a few people wishing to follow the demo seem to have issues getting the LLVM build working due to one reason or another (mostly due to versioning issues it seems). I've removed the version manifest and the instructions for that from the demo.

PR #135, opened 2022-09-06
3.8 instead of '3.8' in main.yml

yundiqian:


.github/workflows/main.yml
```diff
@@ -23,7 +23,7 @@ jobs:
             _ver="['3.8', '3.9', '3.10']"
             _cache="1"
           else 
-            _ver="['3.8']"
+            _ver="[3.8]"
```
mtrofin, line 26: can you put a comment above it that it should stay that way so the `arc` scenario works?

PR #137, opened 2022-09-08
Pass serialized policies to workers

mtrofin:
This sets up the trainer to work with potentially remote workers by passing a serialized object to them. Each worker, then, serializes the policy locally, after which data collection continues as usual.

The main inefficiency is that we save the policy per request handler, and even if the training is local. Policies are small objects, however (at least currently), so this shouldn't be cause of concern. Upon measuring this for 100 and then 500 modules, with 128 workers:

- 100 modules: ~7% regression in the compile time component of training
- 500 modules: ~9% improvement

At this point, I'd take neither regression nor improvement as meaningful, and since, additionally, compilation time is about 1/10 of overall time, the complexity tradeoff of a design optimizing for the local case is probably not worth it.

### DIFF
--- a/compiler_opt/rl/compilation_runner.py
+++ b/compiler_opt/rl/compilation_runner.py
@@ -26,7 +26,8 @@ from typing import Dict, List, Optional, Tuple
 
 from absl import flags
 from compiler_opt.distributed.worker import Worker, WorkerFuture
-from compiler_opt.rl import constant, policy_saver
+from compiler_opt.rl import constant
+from compiler_opt.rl import policy_saver
 from compiler_opt.rl import corpus
 import tensorflow as tf
 
@@ -47,38 +48,6 @@ def _calculate_reward(policy: float, baseline: float) -> float:
 class RewardStat:
   default_reward: float
   moving_average_reward: float
-
-
-@dataclasses.dataclass(frozen=True)
-class Policy:
-  """Serialized mlgo policy, used to pass a policy to workers.
-
-  A policy has 2 components, both being file contents:
-    - the content of the output_spec.json file;
-    - the content of the tflite policy.
-  """
-
-  output_spec: bytes
-  policy: bytes
-
-  def to_filesystem(self, location: str):
-    os.makedirs(location, exist_ok=True)
-    output_sig = os.path.join(location, policy_saver.OUTPUT_SIGNATURE)
-    policy_path = os.path.join(location, policy_saver.TFLITE_MODEL_NAME)
-    with tf.io.gfile.GFile(output_sig, mode='wb') as f:
-      f.write(self.output_spec)
-    with tf.io.gfile.GFile(policy_path, mode='wb') as f:
-      f.write(self.policy)
-
-  @staticmethod
-  def from_filesystem(location: str):
-    output_sig = os.path.join(location, policy_saver.OUTPUT_SIGNATURE)
-    policy_path = os.path.join(location, policy_saver.TFLITE_MODEL_NAME)
-    with tf.io.gfile.GFile(output_sig, mode='rb') as f:
-      output_spec = f.read()
-    with tf.io.gfile.GFile(policy_path, mode='rb') as f:
-      policy = f.read()
-    return Policy(output_spec=output_spec, policy=policy)
 
 
 class DataClassJSONEncoder(json.JSONEncoder):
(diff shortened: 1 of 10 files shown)


compiler_opt/rl/compilation_runner.py
```diff
@@ -20,12 +20,13 @@
...
 import signal
 import subprocess
+import tempfile
 import threading
 from typing import Dict, List, Optional, Tuple
 
 from absl import flags
 from compiler_opt.distributed.worker import Worker, WorkerFuture
-from compiler_opt.rl import constant
+from compiler_opt.rl import constant, policy_saver
```
yundiqian, line 29: separate imports to two lines

compiler_opt/rl/compilation_runner.py
```diff
@@ -48,6 +49,38 @@ class RewardStat:
   moving_average_reward: float
 
 
+@dataclasses.dataclass(frozen=True)
+class Policy:
```
yundiqian, line 53: move it to policy_saver.py?

compiler_opt/rl/compilation_runner.py
```diff
@@ -48,6 +49,38 @@ class RewardStat:
...
+  """Serialized mlgo policy, used to pass a policy to workers.
+
+  A policy has 2 components, both being file contents:
```
yundiqian, line 56: add an usage example in docstring

compiler_opt/rl/compilation_runner.py
```diff
@@ -48,6 +49,38 @@ class RewardStat:
...
+
+  @staticmethod
+  def from_filesystem(location: str):
+    output_sig = os.path.join(location, policy_saver.OUTPUT_SIGNATURE)
+    policy_path = os.path.join(location, policy_saver.TFLITE_MODEL_NAME)
+    with tf.io.gfile.GFile(output_sig, mode='rb') as f:
+      output_spec = f.read()
+    with tf.io.gfile.GFile(policy_path, mode='rb') as f:
+      policy = f.read()
+    return Policy(output_spec=output_spec, policy=policy)
```
yundiqian, line 81: do you want to define an object or just a bytes (like what we are doing today) for it? the Policy object adds more additional complexity when you passing it through rpcs in the distributed case

compiler_opt/rl/compilation_runner.py
mtrofin: done

compiler_opt/rl/compilation_runner.py
mtrofin: done

compiler_opt/rl/compilation_runner.py
mtrofin: done

compiler_opt/rl/compilation_runner.py
mtrofin: What complexity - there's no difference between how this object is sent in the local-multiprocess case or distributed case.

yundiqian: How does the trainer and the server talk with each other in the new distributed case? not in RPCs?

mtrofin: That's handled transparently. The objects, either way (inter-process, local or distributed) are serialized/deserialized the same way under the hood.

yundiqian: i mean is the new distributed framework able to handle the serialization/deserialization of a python object automatically? In our current grpc setup, it cannot --- training in python and data collection in C++, even if the data collection is in python, we need to do the serialization/deserialization of the python object ourselves then we can pass it through rpc request

mtrofin: The short answer is yes - that's part of the value, we don't need to deal with any of that gRPC, protobufs, any of that. Any pickleable object works.

Basically all the code currently written doesn't need to change. We just need to de-localize the source files (the module .bc/thinlto - a subsequent patch) and that's it.

compiler_opt/rl/policy_saver.py
```diff
@@ -110,6 +111,46 @@ def convert_mlgo_model(mlgo_model_dir: str, tflite_model_dir: str):
   tf.io.gfile.copy(src_json, dest_json)
 
 
+@dataclasses.dataclass(frozen=True)
+class Policy:
```
yundiqian, line 115: add a unit test?

yundiqian (approved):
otherwise lgtm

PR #103, opened 2022-08-11
enable gin bindings for compilation flags

kshiteejm:
For Issue #102

PR #142, opened 2022-09-19
delete '-fthinlto-index' in delete_flags

yundiqian:
due to https://github.com/google/ml-compiler-opt/commit/fc26cfcdff7941905cf98e7e67c5a62c5841d201


kshiteejm (commented):
LGTM.

compiler_opt/rl/problem_configuration.py
```diff
@@ -113,5 +113,5 @@ def flags_to_add(self) -> Tuple[str, ...]:
   # List of flags to remove from clang compilation command. The flag names
   # should match the actual flags provided to clang.'
   def flags_to_delete(self) -> Tuple[str, ...]:
-    return ('-split-dwarf-file', '-split-dwarf-output', '-fthinlto-index',
-            '-fprofile-sample-use', '-fprofile-remapping-file')
+    return ('-split-dwarf-file', '-split-dwarf-output', '-fprofile-sample-use',
```
mtrofin, line 116: Shouldn't `-fprofile-sample-use` and `-fprofile-remapping-file`  also be removed?

yundiqian: i'm not sure, why shall we remove them?

kshiteejm: I think they should be there in delete_flags for default case (since they're not used?), but have to be removed from delete_flags and moved to replace_flags if we plan to use those flags for use-cases such as AFDO profile reinjection. I've done that in https://github.com/google/ml-compiler-opt/pull/103/files. PTAL.

mtrofin: I left a comment on #103. We don't want to treat profile flags as delete by default. That's the whole point of modeling them as "replace": if present, and deleted, compilation works but the outcome is not what we want, and it's very hard to identify this as the root cause.

Debug info flags, in contrast, just don't matter to us right now in any scenario. Safe to treat as "delete" by default.

PR #144, opened 2022-09-21
move DataClassJSONEncoder to a more common place for future reuse 

yundiqian:


compiler_opt/rl/constant.py
```diff
@@ -33,3 +35,11 @@ class AgentName(enum.Enum):
   BEHAVIORAL_CLONE = 0
   DQN = 1
   PPO = 2
+
+
+class DataClassJSONEncoder(json.JSONEncoder):
```
mtrofin, line 40: I get why not `compilation_runner`, but `constant` doesn't feel that great either. I'm assuming it's the best out of all the worst? Maybe put a TODO on it to say it may need a better place in the future, with a issue tagged on - so if later we end up with some sort of 'converters' or something like that, we know we didn't have much of a reason for it to be in `constant.py` either.

mtrofin:
lgtm, a nit.

compiler_opt/rl/constant.py
yundiqian: how about utils.py?

mtrofin: 'utils' says very little :) (i.e. I believe it's an anti-pattern - it's a place where we risk putting everything - because what's _not_ util :) )

might as well make a `converters.py` - would be equally populated by one thing now, and avoids the kitchen drawer problem.

(some [light reading](https://dev.to/dvddpl/utils-files-are-not-so-useful-and-helper-classes-are-not-so-helpful-1kfn) illustrating this)

yundiqian: converters may be too specific, i cannot think of other use cases for now...

let's then put it in constant.py for now, we can refactor in the future if needed

PR #153, opened 2022-09-26
add the support for dumping best trajectories during training

yundiqian:
add the support for dumping best trajectories during training for later use. It finds the most optimal action list for individual modules/functions

the naming of policy_rewards, when put together with rewards in the CompilationResult is confusing. will have a full codebase refactor to refactor {default|policy}_reward to {|}_cost after #31 is done


mtrofin:
You can reference #31 instead of "migration to executor", since the latter is an internal effort.

yundiqian:
> 

done

compiler_opt/rl/local_data_collector.py
```diff
@@ -118,7 +121,7 @@ def collect_data(
     """Collect data for a given policy.
 
     Args:
-      policy_path: the path to the policy directory to collect data with.
+      policy: an policy_saver.Policy object to collect data with.
```
mtrofin, line 124: nit: typo (a policy_saver...)

mtrofin:
lgtm. (The naming is now confusing indeed, ack on the future refactoring)

PR #156, opened 2022-09-28
Update the Dockerfiles to use a base image supported by Fuchsia infra.

zeroomega:
The current base image, Ubuntu 22.04LTS has a much newer kernel and glibc. Fuchsia infra builders is running into issues to support it. This patch changes the base image to debian bullseye which aligns with what infra builders use.

This change will drops the python revision from 3.10 to 3.9.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:22.04
+FROM debian:bullseye
 RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y \
     python-is-python3 \
     python3 \


mtrofin:
@boomanaiden154 any reason we used ubuntu 22.04 originally?

(merging to unblock Fuchsia in the meantime)

boomanaiden154:
There wasn't any specific reason I moved from 20.04 to 22.04 originally. I had just moved to using 22.04 as my base image locally, and it fixed some versioning related issues in the experimental development image, and to keep everything in sync I changed the base image of the main `Dockerfile` to `ubuntu:22.04` as well. When making the change I made the assumption that someone would bring it up in code review if this would cause problems for Fuchsia, but in hindsight that was not a good assumption. I'll make sure to ping appropriate reviewers on the Fuchsia side when adjusting anything in regards to the main `Dockerfile` in the future.

mtrofin:
No problem - we just tried it now there, and became aware of this issue; just wanted to check if there was any hard reason. Thanks for taking a look!

PR #160, opened 2022-10-05
Use lower case spec name for future compatibility

faizan-m:


yundiqian:
should be a no-op change to the result

PR #164, opened 2022-10-10
Added patch for extracting corpora from Chromium with ThinLTO

boomanaiden154:
Should be working in upstream Chromium once [this](https://chromium-review.googlesource.com/c/chromium/src/+/3941232) CL gets merged as that toolchain roll will grab [my commit](https://github.com/llvm/llvm-project/commit/f741815ddb49e3ed26de327a51ccf60f66f21d46) that fixes some issues related to thin archives in these builds.


boomanaiden154:
Looks like that CL was superseded by [this](https://source.chromium.org/chromium/chromium/src/+/00694805065ec15fceed25ba2b3f89f852fba372) commit, so the patch in upstream LLD should be in the Chromium toolchain now meaning that this patch will work without an external head-of-tree LLVM toolchain.

mtrofin:
LGTM

mtrofin (approved):

PR #166, opened 2022-10-13
Explicitly pass the worker's gin config

mtrofin:
If the local worker manager's processes aren't forked (which depends on what the default for `multiprocessing.get_context` is) the gin bindings won't be available. So we just pass them explicitly.

### DIFF
--- a/compiler_opt/distributed/local/local_worker_manager.py
+++ b/compiler_opt/distributed/local/local_worker_manager.py
@@ -245,9 +245,9 @@ class LocalWorkerPoolManager(AbstractContextManager):
                *args, **kwargs):
     if not count:
       count = multiprocessing.get_context().cpu_count()
-    final_args = worker.get_full_worker_args(worker_class, kwargs)
+    final_kwargs = worker.get_full_worker_args(worker_class, kwargs)
     self._stubs = [
-        _make_stub(worker_class, *args, **final_args) for _ in range(count)
+        _make_stub(worker_class, *args, **final_kwargs) for _ in range(count)
     ]
 
   def __enter__(self) -> worker.FixedWorkerPool:
--- a/compiler_opt/distributed/local/local_worker_manager.py
+++ b/compiler_opt/distributed/local/local_worker_manager.py
@@ -36,7 +36,7 @@ import threading
 
 from absl import logging
 # pylint: disable=unused-import
-from compiler_opt.distributed.worker import Worker, FixedWorkerPool
+from compiler_opt.distributed import worker
 
 from contextlib import AbstractContextManager
 from multiprocessing import connection
@@ -59,8 +59,8 @@ class TaskResult:
   value: Any
 
 
-def _run_impl(pipe: connection.Connection, worker_class: 'type[Worker]', *args,
-              **kwargs):
+def _run_impl(pipe: connection.Connection, worker_class: 'type[worker.Worker]',
+              *args, **kwargs):
   """Worker process entrypoint."""
 
   # A setting of 1 does not inhibit the while loop below from running since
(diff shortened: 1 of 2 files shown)


compiler_opt/distributed/local/local_worker_manager.py
```diff
@@ -245,8 +248,18 @@ def __init__(self, worker_class: 'type[Worker]', count: Optional[int], *args,
...
+    except:  # pylint: disable=bare-except
+      pass
+    # Issue #38
+    if sys.version_info.minor >= 9:
+      final_args = kwargs | gin_config
+    else:
+      final_args = {**kwargs, **gin_config}
     self._stubs = [
-        _make_stub(worker_class, *args, **kwargs) for _ in range(count)
+        _make_stub(worker_class, *args, **final_args) for _ in range(count)
```
yundiqian, line 262: Jacob also encountered this issue. One possible solution is to switch to the multiprocessing module in TFAgents that can pass gin through. but he encountered some issues after the switch. Not sure whether he solved it or now.

mtrofin: This solves the problem uniformly for any distribution model, including Dask, or our internal distro framework, for example.

mtrofin: (i.e. next step I'd move this in a common place)

jacob-hegna: this would fix the issue I was running into. as for switching the multiprocessing modules, I think it's probably better to handle the issue by explicitly passing the gin flags rather than depend on subtle interactions between gin and the multiprocessing module.

compiler_opt/distributed/worker.py
```diff
@@ -86,3 +89,21 @@ def get_exception(worker_future: WorkerFuture) -> Optional[Exception]:
...
+  """Get the union of given kwargs and gin config.
+
+  This allows the worker hosting process be set up differently from the training
+  process - e.g. no need to initialize gin variables there, for example.
+  """
+  gin_config = {}
+  try:
+    gin_config = gin.get_bindings(worker_class)
+  except:  # pylint: disable=bare-except
+    pass
```
yundiqian, line 104: what exception can it throw? and shall we silently ignore the exception?

compiler_opt/distributed/local/local_worker_manager.py
```diff
@@ -241,16 +241,17 @@ def __dir__(self):
 class LocalWorkerPoolManager(AbstractContextManager):
   """A pool of workers hosted on the local machines, each in its own process."""
 
-  def __init__(self, worker_class: 'type[Worker]', count: Optional[int], *args,
-               **kwargs):
+  def __init__(self, worker_class: 'type[worker.Worker]', count: Optional[int],
+               *args, **kwargs):
     if not count:
       count = multiprocessing.get_context().cpu_count()
+    final_args = worker.get_full_worker_args(worker_class, kwargs)
```
yundiqian, line 248: nit: final_kwargs

compiler_opt/distributed/worker.py
mtrofin: left a comment. ValueError, funny enough, because it's not possible to figure out if the object even has a gin config without trying.

compiler_opt/distributed/local/local_worker_manager.py
mtrofin: done

compiler_opt/distributed/worker.py
yundiqian: I'm a bit concerned about the silent ignorance of all exceptions, e.g., spelling mistakes, etc. Ideally we want to separate the exceptions we want to capture and the the exceptions we meant to ignore.

The bottom line here is to capture ValueError instead of ignoring all exceptions

mtrofin: Done. I wish they threw something more specific, because "ValueError" is very well all those examples you gave, too (they should be caught by the typechecker, but surely there's others in the same vein)

Maybe I'll contribute a patch to gin.

yundiqian (approved):

PR #172, opened 2022-10-27
Convenience: allow specifying the corpus description json file

mtrofin:
I found it useful sometimes to have variations of the corpus description - e.g. remove swaths of modules. This patch allows multiple descriptions to co-exist under a corpus directory, and select the desired one transparently.

Renamed the ctor argument since "data path" is more of a directory than a file, "location" covers both cases.

### DIFF
--- a/compiler_opt/rl/corpus.py
+++ b/compiler_opt/rl/corpus.py
@@ -32,6 +32,8 @@ from compiler_opt.rl import constant
 # command line, where all the flags reference existing, local files.
 FullyQualifiedCmdLine = Tuple[str, ...]
 
+DEFAULT_CORPUS_DESCRIPTION_FILENAME = 'corpus_description.json'
+
 
 def _apply_cmdline_filters(
     orig_options: Tuple[str, ...],
@@ -185,7 +187,7 @@ class SamplerBucketRoundRobin(Sampler):
 class Corpus:
   """Represents a corpus.
 
-  A corpus is created from a corpus_description.json file, produced by
+  A corpus is created from a corpus description json file, produced by
   extract_ir.py (for example).
 
   To use the corpus:
@@ -225,7 +227,7 @@ class Corpus:
 
   def __init__(self,
                *,
-               data_path: str,
+               location: str,
                module_filter: Optional[Callable[[str], bool]] = None,
                additional_flags: Tuple[str, ...] = (),
                delete_flags: Tuple[str, ...] = (),
@@ -238,7 +240,10 @@ class Corpus:
     output) and validated.
 
     Args:
-      data_path: corpus directory.
+      location: either the path to the corpus description json file in a corpus
+        directory, or just the corpus directory, case in which
+        `DEFAULT_CORPUS_DESCRIPTION_FILENAME` will be assumed to be available
+        in that directory and used as corpus description.
       additional_flags: list of flags to append to the command line
       delete_flags: list of flags to remove (both `-flag=<value` and
         `-flag <value>` are supported).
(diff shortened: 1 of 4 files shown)


compiler_opt/rl/corpus.py
```diff
@@ -238,7 +240,10 @@ def __init__(self,
     output) and validated.
 
     Args:
-      data_path: corpus directory.
+      location: either the path to the corpus description json file in a corpus
```
yundiqian, line 243: Is it better that we only allow the input being the corpus_description path? The only concern is that it adds some migration cost...

mtrofin: I think that'd be better, but to avoid migration pain, we could stage it: this patch first; then warn about deprecation; then deprecate 

we can skip the 2nd step because we don't have a release model and so when folks would notice the change is out of our control; but can work with e.g. Fuchsia to update their scripts.

compiler_opt/rl/corpus.py
```diff
@@ -32,6 +32,8 @@
 # command line, where all the flags reference existing, local files.
 FullyQualifiedCmdLine = Tuple[str, ...]
 
+DEFAULT_CORPUS_DESCRIPTION_FILENAME = 'corpus_description.json'
```
yundiqian, line 35: _DEFAULT_CORPUS_...

compiler_opt/rl/corpus.py
```diff
@@ -225,7 +227,7 @@ class ReplaceContext:
 
   def __init__(self,
                *,
-               data_path: str,
+               location: str,
```
yundiqian, line 230: location is a bit confusing, seems that data_path is a more informative name

compiler_opt/rl/corpus.py
yundiqian: the migration plan sgtm!

compiler_opt/rl/corpus.py
yundiqian: _DEFAULT_...

yundiqian: I mean adding underscore('_') before DEFAULT_...

github believes that '_'DEFAULT_... means italic the word following the '_'

compiler_opt/rl/corpus.py
mtrofin: I'm fine to change to anything, but did you read the commit message (because I'm making the opposite argument there).

yundiqian: oh i missed that part. 

data_path sounds to me that can represents both file path or dir path. My major concern about the naming of location is: 1) it does not describe it is the location of what so can be a bit confusing, data_location may be more descriptive in that sense; 2) the flag name is data_path now, having different names increases the burden a little bit.

mtrofin: How about `corpus_description`, since we'd subsequently go with deprecating dir-based value anyway? (probably subsequently worth updating the flag, too)

yundiqian: The issue with corpus_description is that it is a bit too specific so if in the future we have more/different corpus formatting we have to do some migration work again. Meanwhile, data_path is a general and reasonable name imo.

I'm fine with either as long as the name is the same as the flag name, so up to you :)

mtrofin: `corpus_path` (and ack on the flag). Less general than "data", reasonably ambiguous-ish to mean either dir or file, and says nothing about description or anything like that. wdyt?

yundiqian: corpus_path sounds great!

Issue #146, opened 2022-09-22
Investigate lengthy data collection in python 3.10


mtrofin:
All else being the same, compiling ~100 modules takes ~4-5s when using python3.9, but blows up to 30s when moving to python 3.10.

mtrofin:
So far, using viztrace, was able to pinpoint protobuf serialization as very likely root cause. The protobuf package is at the same version, though, so looking further.

mtrofin:
It seems that updating to `protobuf==3.19.5` addresses that. Would prefer postpone changing the version in `requirements.txt` to when we bump the tf version, and this is another reason to move away from protobuf altogether.

boomanaiden154:
I believe the protobuf version has been updated to `3.19.5` in the version of Tensorflow that we're already using (or the version is just set to be `> 3.x.x && < 4.0.0`. When I regenerated the lockfile in #128, it automatically grabbed `3.19.5` as when we upgraded to Tensorflow nightly originally, we didn't really touch any of the dependency versions (#118).

mtrofin:
huh. so I must have quite a messed up local setup.

boomanaiden154:
It's not your local setup. The protobuf version in `requirements.txt` is still set to the lower version.

boomanaiden154:
Testing after #189:
Default trace over an LLVM corpus, default settings (Python 3.8.10):
```
real	1m43.543s
user	120m39.134s
sys	4m22.878s
```
Warmstart:
```
real	1m32.171s
user	4m54.158s
sys	3m45.621s
```
PPO training, default settings (except setting `num_policy_iterations`):
```
real	6m42.992s
user	219m37.325s
sys	40m12.776s
```
Default trace over an LLVM corpus, default settings (Python 3.10.6):
```
real	1m41.111s
user	121m11.636s
sys	2m33.097s
```
Warmstart:
```
real	1m30.252s
user	5m6.882s
sys	3m32.267s
```
PPO training, default settings (except setting `num_policy_iterations`):
```
real	6m32.612s
user	199m24.214s
sys	34m36.932s
```
While these runs aren't perfectly controlled to just the data collection step (although the default trace runs should come close), they do show that the performance is pretty comparable between 3.10 and previous versions now. I'm seeing similar times for module compilation between both versions and about the performance I'd "expect" given timings/performance that I've seen before.

[...]

PR #174, opened 2022-10-28
Use the official Python image as base image

petrhosek:
This gives more precise control over the Python version.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,9 @@
-FROM debian:bullseye
-RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y \
-    python-is-python3 \
-    python3 \
-    python3-pip \
-    && rm -rf /var/lib/apt/lists/*
+FROM python:3.9
+
 COPY . /ml-compiler-opt
-RUN python3 -m pip install -r /ml-compiler-opt/requirements.txt
+RUN pip install --no-cache-dir -r /ml-compiler-opt/requirements.txt
+
+WORKDIR /ml-compiler-opt/compiler_opt/tools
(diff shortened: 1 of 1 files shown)


petrhosek:
@zeroomega This is working for me locally but we should also test this in Fuchsia's infrastructure.

boomanaiden154 (commented):
I'd caution against using Python 3.10 for now. See #146. This should be fixed once we go through with #128 (due to the dependency updates there), but for now we've being doing most stuff on Python 3.8/3.9.

zeroomega:
Once this lands I can manually test it on Fuchsia infra builders.

Dockerfile
```diff
@@ -1,10 +1,9 @@
...
-RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y \
-    python-is-python3 \
-    python3 \
-    python3-pip \
-    && rm -rf /var/lib/apt/lists/*
+FROM python:3.9
+
 COPY . /ml-compiler-opt
-RUN python3 -m pip install -r /ml-compiler-opt/requirements.txt
+RUN pip install --no-cache-dir -r /ml-compiler-opt/requirements.txt
```
boomanaiden154, line 4: What's the rationale here for using `--no-cache-dir`?

petrhosek: The cache is used to store the installation files(`.whl`, etc.) of the modules that you install through pip and the source files (`.tar.gz`, etc.) neither of which are needed inside the container, so this helps to reduce the size of the layer.

boomanaiden154: Makes perfect sense. Was thinking of the wrong file when I saw the line.

boomanaiden154 (approved):

PR #175, opened 2022-10-28
Support Pipenv

petrhosek:
Pipenv combines Pipfile, pip, and virtualenv into one single command. This allows specifying only the direct dependencies, and relying on Pipenv for generating the lockfile and install packages.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -1,0 +1,22 @@
+[[source]]
+url = "https://pypi.python.org/simple"
+verify_ssl = true
+name = "pypi"
+
+[packages]
+absl-py = "1.0.0"
+gin = "0.1.6"
+joblib = "1.1.0"
+psutil = "5.9.0"
+tf-agents = "0.15.0rc0"
+tensorflow = "2.11.0rc1"
+
+[dev-packages]
+pylint = "2.14.1"
+pytest = "7.1.2"
+pytype = "2022.06.06"
+yapf = "0.32.0"
+
+[ci]
+joblib = "1.1.0"
+shap = "0.41.0"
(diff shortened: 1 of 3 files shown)


petrhosek:
I'm currently seeing the following error when running `run_tests.sh` inside the container:
```
DeprecationWarning: distutils Version classes are deprecated. Use packaging.version instead.
```
This still needs to be addressed.

boomanaiden154:
This looks a lot better than what we're doing currently and is definitely a much better option than what I had proposed in #128. I'm assuming the move from `tf-nightly` to `tensorflow==2.11.0rc1` is because the changes that we were depending upon in nightly before have now made the 2.11.0 branch?

There should probably be some minor follow ups on this (ie updating documentation and maybe some documentation on updating dependencies), but I can take care of those later and we can get this in so that you can get unblocked.

boomanaiden154:
> I'm currently seeing the following error when running `run_tests.sh` inside the container:
> 
> ```
> DeprecationWarning: distutils Version classes are deprecated. Use packaging.version instead.
> ```
> 
> This still needs to be addressed.

The current protocol for deprecation warnings in downstream dependencies is to open an Issue documenting the warning and then add a line in `pytest.ini` to ignore it.

mtrofin:
Can you please update the CI to use this, too - and then can we remove the requirements.txt family of files?

petrhosek:
> Can you please update the CI to use this, too - and then can we remove the requirements.txt family of files?

Where is the CI configuration?

boomanaiden154:
> Where is the CI configuration?

`./github/workflows/main.yml`

boomanaiden154:
#189 finished this transition since Petr is OOO for the rest of the year.

PR #176, opened 2022-11-03
Prepare for distributed training infrastructure.

jacob-hegna:
Distributed training has slightly different requirements than our current setup. This commit includes some small changes to prepare for the distributed training commit.

Relevant changes:
 - add ability to observe compilation results within each worker (this will be used to efficiently send experiences to reverb). This is intended for users who need to observe compilations while they are in the distributed worker pool, rather than after they have been coalesced in the collection script.

 - add a sequence example dataset creator which doesn't batch examples (reverb does not handle batched examples).

 - slightly change the parsing logic to handle distributed PPO evaluation and training. In particular, distributed training requires that compute_value_and_advantage_in_train=False, which changes the policy_info spec that we require. Similarly, when evaluating the policy we convert the collect policy parameters stored on reverb to a greedy policy, which requires slightly changing how we parse data. Nothing changes in the "default" training case.

 - add a "model_id" parameter for collect_data() which is used to correctly tag experiences with the model which collected them during distributed training. This is used to ensure that the PPO training remains on-policy in the distributed setting.

 - change the ConstantValueNetwork to determine its output shape dynamically based on the input shape.

### DIFF
--- a/compiler_opt/rl/compilation_runner.py
+++ b/compiler_opt/rl/compilation_runner.py
@@ -311,7 +311,8 @@ class CompilationRunner(Worker):
                launcher_path: Optional[str] = None,
                moving_average_decay_rate: float = 1,
                compilation_timeout=None,
-               create_observer_fns: Optional[List[Callable]] = None):
+               create_observer_fns: Optional[List[Callable[
+                   [], CompilationResultObserver]]] = None):
     """Initialization of CompilationRunner class.
 
     Args:
--- a/compiler_opt/rl/compilation_runner.py
+++ b/compiler_opt/rl/compilation_runner.py
@@ -332,8 +332,8 @@ class CompilationRunner(Worker):
     self._compilation_timeout = (
         compilation_timeout or _COMPILATION_TIMEOUT.value)
     self._cancellation_manager = WorkerCancellationManager()
-    self._observers = [f() for f in create_observer_fns
-                      ] if create_observer_fns else []
+    self._observers = ([f() for f in create_observer_fns]
+                       if create_observer_fns else [])
 
   # re-allow the cancellation manager accept work.
   def enable(self):
--- a/compiler_opt/rl/compilation_runner.py
+++ b/compiler_opt/rl/compilation_runner.py
@@ -21,7 +21,7 @@ import signal
 import subprocess
 import tempfile
 import threading
-from typing import Dict, List, Optional, Tuple
+from typing import Callable, Dict, List, Optional, Tuple
 
 from absl import flags
 from absl import logging
@@ -276,6 +276,27 @@ class CompilationRunnerStub(metaclass=abc.ABCMeta):
     raise NotImplementedError()
 
 
+class CompilationResultObserver(metaclass=abc.ABCMeta):
+  """Abstract base class used to observe compilation results.
+
+  This is indended for users who need to observe compilations while they are in
+  the distributed worker pool, rather than after they have been coalesced in
+  the collection script.
+  """
+
+  @abc.abstractmethod
+  def observe(self, result: CompilationResult) -> None:
+    """Observe a compilation result.
+
+    Note that this will be executed on the worker in the pool rather than on
+    the coordinator.
+
+    Args:
+      result: the compilation result to observe
+    """
+    pass
+
+
 class CompilationRunner(Worker):
   """Base class for collecting compilation data."""
 
@@ -289,13 +310,20 @@ class CompilationRunner(Worker):
                clang_path: Optional[str] = None,
                launcher_path: Optional[str] = None,
                moving_average_decay_rate: float = 1,
-               compilation_timeout=None):
+               compilation_timeout=None,
+               create_observer_fns: Optional[List[Callable]] = None):
     """Initialization of CompilationRunner class.
 
     Args:
       clang_path: path to the clang binary.
       launcher_path: path to the launcher binary.
       moving_average_decay_rate: moving average decay rate during training.
+      create_observer_fns: list of callables which are used to create
+        CompilationResultObserver objects. We pass the create callables instead
+        of the actual objects because the callables are "just code" and likely
+        able to be pickled/sent to the workers, whereas the observers might
+        contain non-picklable attributes, such as server connections. It is
+        typed as Optional[List[]] due to W0102 dangerous-default-value.
     """
     self._clang_path = clang_path
     self._launcher_path = launcher_path
@@ -304,6 +332,8 @@ class CompilationRunner(Worker):
     self._compilation_timeout = (
         compilation_timeout or _COMPILATION_TIMEOUT.value)
     self._cancellation_manager = WorkerCancellationManager()
+    self._observers = [f() for f in create_observer_fns
+                      ] if create_observer_fns else []
 
   # re-allow the cancellation manager accept work.
   def enable(self):
@@ -388,13 +418,18 @@ class CompilationRunner(Worker):
       policy_rewards.append(policy_reward)
       keys.append(k)
 
-    return CompilationResult(
+    result = CompilationResult(
         sequence_examples=sequence_example_list,
         reward_stats=reward_stat,
         rewards=rewards,
         policy_rewards=policy_rewards,
         keys=keys,
         model_id=model_id)
+
+    for observer in self._observers:
+      observer.observe(result)
+
+    return result
 
   def compile_fn(
       self, command_line: corpus.FullyQualifiedCmdLine, tf_policy_path: str,
--- a/compiler_opt/rl/constant_value_network.py
+++ b/compiler_opt/rl/constant_value_network.py
@@ -47,6 +47,6 @@ class ConstantValueNetwork(network.Network):
 
   def call(self, inputs, step_type=None, network_state=(), training=False):
     _ = (step_type, training)
-    shape = nest_utils.get_outer_array_shape(inputs, self._input_tensor_spec)
-    return tf.constant(
-        self._constant_output_val, tf.float32, shape=shape), network_state
+    shape = nest_utils.get_outer_shape(inputs, self._input_tensor_spec)
+    constant = tf.constant(self._constant_output_val, tf.float32)
+    return tf.fill(shape, constant), network_state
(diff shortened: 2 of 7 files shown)


compiler_opt/rl/compilation_runner.py
```diff
@@ -260,8 +263,8 @@ def collect_data(
       self,
       loaded_module_spec: corpus.LoadedModuleSpec,
       policy: Optional[policy_saver.Policy] = None,
-      reward_stat: Optional[Dict[str, RewardStat]] = None
-  ) -> WorkerFuture[CompilationResult]:
+      reward_stat: Optional[Dict[str, RewardStat]] = None,
+      model_id: Optional[int] = None) -> WorkerFuture[CompilationResult]:
```
mtrofin, line 267: can you peel `model_id` change out first, make it mandatory, and use it in the name of the temporary policy file in `collect_data` - pass the iteration ID or something from ppo to populate it

reason: right now, if a compilation fails, we can dump out the clang command line (quiet=False - or just log it out, whcih we don't but we could). The input files have a canonical path that makes them easily identifiable back in the corpus. The policy, not so much. This would fix that.

jacob-hegna: Done (#178). The model_id attribute is optional in the CompilationResult because the API for CompilationRunner allows compiling without a policy, but in the data collectors the model_id parameter is required.

mtrofin:
Can you copy in the description of the patch the blurb in `compilation_runner.py` that explains the motivation for the observer (i.e. that it's so it doesn't need to be coalesced by the trainer)

Why is that, btw? The expectation is that the sequence examples would be too large when needing distributed training?

LGTM otherwise, but the changes are mostly in training logic, so please wait for @yundiqian 's review.

jacob-hegna:
> Can you copy in the description of the patch the blurb in `compilation_runner.py` that explains the motivation for the observer (i.e. that it's so it doesn't need to be coalesced by the trainer)
> 
> Why is that, btw? The expectation is that the sequence examples would be too large when needing distributed training?

In distributed training we have separate servers for experience collection and training. They are connected by reverb: in the naive approach where we iterate through the compilation results after they have been collected from the pool and send each on to reverb, we are bottlenecked by that loop (some combination of python being slow at looping, and also doing thousands of synchronous network calls). In this new setup, we hoist that loop to each individual worker in the pool so that sending experiences to reverb can happen in parallel. This dramatically speeds up training in the distributed setting.

compiler_opt/rl/constant_value_network.py
```diff
@@ -47,6 +47,6 @@ def __init__(self, input_tensor_spec, constant_output_val=0, name=None):
 
   def call(self, inputs, step_type=None, network_state=(), training=False):
     _ = (step_type, training)
-    shape = nest_utils.get_outer_array_shape(inputs, self._input_tensor_spec)
-    return tf.constant(
-        self._constant_output_val, tf.float32, shape=shape), network_state
+    shape = nest_utils.get_outer_shape(inputs, self._input_tensor_spec)
```
yundiqian, line 50: does it change the code behavior?

compiler_opt/rl/data_reader.py
```diff
@@ -24,16 +24,21 @@
...
-      return {
+      parsing_dict = {
           'CategoricalProjectionNetwork_logits':
               tf.io.FixedLenSequenceFeature(
                   shape=(action_spec.maximum - action_spec.minimum + 1),
-                  dtype=tf.float32)
+                  dtype=tf.float32),
       }
+      if agent_name == constant.AgentName.PPO_DISTRIBUTED:
+        parsing_dict['value_prediction'] = tf.io.FixedLenSequenceFeature(
```
yundiqian, line 39: this causes some code discrepancy between distributed and non-distributed training that we'd want to avoid. 

is it possible to remove it?

compiler_opt/rl/data_reader.py
```diff
@@ -45,7 +50,7 @@ def _get_policy_info_parsing_dict(agent_name, action_spec):
 
 
 def _process_parsed_sequence_and_get_policy_info(parsed_sequence, agent_name,
-                                                 action_spec):
+                                                 action_spec, is_greedy):
```
yundiqian, line 53: what's the use case for is_greedy == True?

compiler_opt/rl/data_reader.py
```diff
@@ -140,27 +150,28 @@ def _parser_fn(serialized_proto):
...
 
 
-def create_sequence_example_dataset_fn(
-    agent_name: constant.AgentName, time_step_spec: types.NestedSpec,
-    action_spec: types.NestedSpec, batch_size: int,
-    train_sequence_length: int) -> Callable[[List[str]], tf.data.Dataset]:
+def create_flat_sequence_example_dataset_fn(agent_name: constant.AgentName,
+                                            time_step_spec: types.NestedSpec,
+                                            action_spec: types.NestedSpec,
+                                            is_greedy: bool = False):
```
yundiqian, line 156: add return type

compiler_opt/rl/constant_value_network.py
jacob-hegna: the output of the network is the same as before (same shape + constant value), however this implementation is potentially slightly slower because it checks the dynamic shape of the input rather than the static shape. I haven't noticed this being an issue though - the policy network is still significantly slower in comparison.

compiler_opt/rl/data_reader.py
jacob-hegna: This is because we set `compute_value_and_advantage_in_train=False` for distributed training. TPUs use mini batches when learning, and that parameter must be False if mini batches are enabled. See: https://github.com/tensorflow/agents/blob/master/tf_agents/agents/ppo/ppo_agent.py#L273

compiler_opt/rl/data_reader.py
jacob-hegna: In distributed training, we have a separate job purely for evaluation and writing stats to tensorboard. That job reads the most recent policy weights from reverb and performs evaluation. Because we are not evaluating using the collect_policy, we don't have the `CategoricalProjectionNetwork_logits` in the output, and so this parsing fails.

To simplify the integration, we could have the distributed learning setup evaluate using the collect_policy, which would make the evaluation a little less faithful but remove this code complexity.

compiler_opt/rl/data_reader.py
jacob-hegna: done.

compiler_opt/rl/data_reader.py
yundiqian: we are using mini batches in the non-distributed training as well, i don't think it's really a hard requirement

compiler_opt/rl/data_reader.py
yundiqian: the trainer naturally does the `evaluation` of the collect policy during training so i don't see a lot of value in setting up an evaluation job to evaluate the collect policy. 

w.r.t. evaluating the policy during training, how much is the performance difference of the policy and the collect policy?

jacob-hegna: The distributed training infrastructure closely follows the example here: https://github.com/tensorflow/agents/tree/master/tf_agents/experimental/distributed

Evaluation is done separately from the training loop to reduce time spent context switching between the host and device (tpu).

I can remove the is_greedy flag for now, though, just to make the integration more simple.

compiler_opt/rl/data_reader.py
jacob-hegna: For distributed training, we are using the tf_agents `PPOLearner` as it is optimized for tpu training (no context switches between device and host, etc). The requirement for `compute_value_and_advantage_in_train=False` comes from this line (https://github.com/tensorflow/agents/blob/10ad6b054bb508adc5a8543ea4849d21f6cc6304/tf_agents/train/ppo_learner.py#L176) which isn't an issue in the non-distributed training as we have our own custom trainer.

compiler_opt/rl/feature_ops.py
```diff
@@ -42,7 +42,8 @@ def build_quantile_map(quantile_file_dir: str):
 
 def discard_fn(obs: types.Float):
```
jacob-hegna, line 43: @yundiqian note the change to this function that I just added - it is similar to the change to ConstantValueNetwork to avoid relying on static tensor shapes (obs.shape) and instead use the dynamic tensor shape. This has a slight functional change as the discarded tensor shapes now have a final dimension of size 1 instead of 0, but the input data is still zeroed out. The only observable difference is a change in the signature of the model after the preprocessing layers.

compiler_opt/rl/compilation_runner.py
```diff
@@ -304,6 +332,8 @@ def __init__(self,
     self._compilation_timeout = (
         compilation_timeout or _COMPILATION_TIMEOUT.value)
     self._cancellation_manager = WorkerCancellationManager()
+    self._observers = [f() for f in create_observer_fns
```
yundiqian, line 335: better indenting to improve the readability

compiler_opt/rl/compilation_runner.py
```diff
@@ -289,13 +310,20 @@ def __init__(self,
                clang_path: Optional[str] = None,
                launcher_path: Optional[str] = None,
                moving_average_decay_rate: float = 1,
-               compilation_timeout=None):
+               compilation_timeout=None,
+               create_observer_fns: Optional[List[Callable]] = None):
```
yundiqian, line 314: pytype the return type for Callable

compiler_opt/rl/data_reader.py
yundiqian: it seems that you patched the original PPOLearner implementation from PR #184 . Is it possible to remove this constraint?

jacob-hegna: I spent some time this morning trying to do this and unfortunately it seems that although it is theoretically possible, it would require a lot of work. It seems like there is an implicit assumption in tf_agents that in the distributed setting, the value computation happens during data collection. I got around this in some spots, but then ran into other issues around the data pipeline. I think the easiest solution would be to include this bit of code in data_reader (I don't think it's too bad, just two extra `if` statements that are clearly intended for the distributed agent)

compiler_opt/rl/compilation_runner.py
jacob-hegna: fixed.

compiler_opt/rl/compilation_runner.py
jacob-hegna: fixed

PR #177, opened 2022-11-03
Bumped setup-python version to fix Github CI deprecation warnings

boomanaiden154:
CI currently is giving a lot of deprecation warnings of the form:
```
The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
```
Due to using an old version of the `setup-python` action. This PR bumps the `setup-python` version to v4 so that the action doesn't use the deprecated path anymore, thus getting rid of the deprecation warning.


mtrofin:
sweet, thanks!

PR #178, opened 2022-11-09
Add a model_id parameter to collect_data.

jacob-hegna:
The model_id parameter allows joining compilation results with the model used during the compilation. This is useful for debugging errors in training as well as for ensuring experience freshness in the distributed training setting.

### DIFF
--- a/compiler_opt/rl/data_collector.py
+++ b/compiler_opt/rl/data_collector.py
@@ -52,12 +52,13 @@ class DataCollector(metaclass=abc.ABCMeta):
 
   @abc.abstractmethod
   def collect_data(
-      self, policy: policy_saver.Policy
+      self, policy: policy_saver.Policy, model_id: int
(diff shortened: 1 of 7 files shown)


compiler_opt/rl/data_collector.py
```diff
@@ -126,7 +127,7 @@ def wait(self, get_num_finished_work):
 
     Args:
       get_num_finished_work: a callable object which returns the amount of
-      finished work.
+        finished work.
```
mtrofin, line 130: spurious?

mtrofin (approved):

compiler_opt/rl/data_collector.py
jacob-hegna: unrelated to the change but my python autoformatter prefers it this way. consistent with the file as well: see the docstring for thresholds on the EarlyExitChecker __init__ method.

PR #184, opened 2022-11-28
Add distributed PPO training.

jacob-hegna:
Distributed training works via four processes:

 - ppo_reverb_server.py
 - ppo_train.py
 - ppo_collect.py
 - ppo_eval.py

ppo_train, ppo_collect, and ppo_eval communicate via ppo_reverb_server. The reverb server has two tables of data: one that holds the most recent model weights, and another that holds training data. The train script fetchs examples from reverb, trains, then uploads new model weights to reverb. The collect script fetches the latest model weights, collects training data, then uploads it to reverb. The eval script fetches the latest model weights and writes evaluation data to tensorboard.

Training is distributed, so each of these processes can live on different servers, and you can have as many collect and eval jobs as you need.

Distributed accelerator training is done via tf.distribute.Strategy. The ppo_train_lib.py script accepts a Strategy as input, and the ppo_train.py script uses the default strategy (cpu training). If you have access to TPUs, you can write an alternate ppo_train.py script that passes a TPUStrategy instead. Similarly, if your server has multiple GPUs, you can use a MirroredStrategy to train on all of them.

### DIFF
--- a/compiler_opt/rl/agent_creators.py
+++ b/compiler_opt/rl/agent_creators.py
@@ -27,6 +27,7 @@ from tf_agents.typing import types
 
 from compiler_opt.rl import constant
 from compiler_opt.rl import constant_value_network
+from compiler_opt.rl.distributed import agent as distributed_ppo_agent
 
 
 def _create_behavioral_cloning_agent(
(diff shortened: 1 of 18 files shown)


jacob-hegna:
The structure of the distributed training infrastructure borrows from other similar projects, as well as from the TFAgents distributed training examples. See:
1) [Circuit training](https://github.com/google-research/circuit_training/tree/main/circuit_training/learning): I reused the structure of their learner + agent after a suggestion from @ebrevdo; before seeing their implementation I kept running in to bugs from outdated documentation.
2) [TFAgents example](https://github.com/tensorflow/agents/tree/master/tf_agents/experimental/distributed).

yundiqian (approved):

PR #186, opened 2022-12-12
Use create_tempdir in corpus_test.py.

jacob-hegna:
Previous, test_braces_in_cmd used a fixed tempdir 'some/temp/dir' which would break the test in some environments. This changes the test to use a directory created from self.create_tempdir(), following the convention of other tests in the file.


mtrofin:
lgtm, thanks!

PR #189, opened 2022-12-14
Switched to dependency management using pipenv

boomanaiden154:
Continuation of #175 since Petr is currently out of the office and the installation of our  current pinned version of `tf-nightly` is broken.

This patch makes a couple main changes
* Switches from using `requirements-*.txt` to using a `Pipfile` and a `Pipfile.lock`.
* Moves CI to using pipenv
* Updates `run_tests.sh` to  use pipenv
* Updates some versions, particularly `dm-reverb-nightly`, `tf-nightly`, and `tf-agents-nightly`. The `tf-*` and `dm-reverb-nightly` packages are supposed to be compatible on the same day, and since we have a needed [patch](https://github.com/tensorflow/agents/commit/af03fbf90aa664f1fd5b72ff7445a782ae2a9edd) that hasn't hit a release or RC yet, I've used the nightly versions. Bumping them to today because the packages do eventually get knocked off PyPI.


Pipfile
```diff
@@ -0,0 +1,23 @@
...
+verify_ssl = true
+name = "pypi"
+
+[packages]
+absl-py = "==1.0.0"
+gin = "==0.1.6"
+joblib = "==1.1.0"
+psutil = "==5.9.0"
+tf-agents-nightly = "==0.15.0dev20221214"
+tf-nightly = "==2.12.0dev20221214"
```
mtrofin, line 12: let's keep this to 2.11 for now.

mtrofin:
Can we have the explicit transitive closure listed in the pipenv.lock, and also enforced that it is actually the transitive closure (pip -r had a way - which we didn't use, but nevertheless - to say "don't auto-pull any other dep")

There are 2 reasons: one is for projects like Fuchsia that want to pull packages from a vetted (by them) location, so knowing what the transitive closure is matters to them (esp. when we bump versions)

Second, package inter-dependencies are often expressed as version ranges (lile >=x.y.z). Given how many packages there are, removing version differences can help reproducing some quirky bug by eliminating at least one large source of differences.

Otherwise lgtm.

mtrofin:
can you also update https://github.com/google/ml-compiler-opt/blob/main/buildbot/buildbot_init.sh#LL116

boomanaiden154:
> Can we have the explicit transitive closure listed in the pipenv.lock, and also enforced that it is actually the transitive closure (pip -r had a way - which we didn't use, but nevertheless - to say "don't auto-pull any other dep")

Using `pipenv sync` installs the packages exactly as specified in `Pipfile.lock` without even touching `Pipfile`, so installing packages that way as I've done in the CI, `run_tests.sh`, and on my local machine should be 100% reproducible. There's an option to get somewhat similar behavior with `pipenv install`, but `pipenv sync` is designed for this task.

> Second, package inter-dependencies are often expressed as version ranges (lile >=x.y.z). Given how many packages there are, removing version differences can help reproducing some quirky bug by eliminating at least one large source of differences.

All of the versions in `Pipfile.lock` should be pinned to exact versions, so after install with `pipenv sync`, there shouldn't be any version differences.

Pipfile
boomanaiden154: The earliest version of `tf-agents-nightly` that contains Jacob's patch should be 20221102, which is `0.15.0` for `tf-agents-nightly` and `2.12.0` for `tf-nightly`. Would it be fine to just have these mismatched (like we're doing now) despite the recommendations from Tensorflow?

jacob-hegna:
Do we need to change the root README.md (in other words, does this change the cmdline to install dependencies)?

Pipfile
mtrofin: ok, let's see w. 2.12 then. Once this lands I'll reboot the AOT buildbots, if they work => all good

mtrofin:
> > Can we have the explicit transitive closure listed in the pipenv.lock, and also enforced that it is actually the transitive closure (pip -r had a way - which we didn't use, but nevertheless - to say "don't auto-pull any other dep")
> 
> Using `pipenv sync` installs the packages exactly as specified in `Pipfile.lock` without even touching `Pipfile`, so installing packages that way as I've done in the CI, `run_tests.sh`, and on my local machine should be 100% reproducible. There's an option to get somewhat similar behavior with `pipenv install`, but `pipenv sync` is designed for this task.
> 
> > Second, package inter-dependencies are often expressed as version ranges (lile >=x.y.z). Given how many packages there are, removing version differences can help reproducing some quirky bug by eliminating at least one large source of differences.
> 
> All of the versions in `Pipfile.lock` should be pinned to exact versions, so after install with `pipenv sync`, there shouldn't be any version differences.

Ah! I misread the diff, I see now, the lock has the transitive closure. Sorry. LGTM

Pipfile
boomanaiden154: Sounds good. Rolling back to `2.11.0dev20221017`.

PR #190, opened 2022-12-15
Update Dockerfile to use pipenv

boomanaiden154:
This updates the `Dockerfile` to use `pipenv` after #189. Also clears the `pipenv` cache (which also clears the pip cache) afterwards since it's a waste couple hundred MB.

Pinging other people on the Fuchsia team who have made changes here before that Github won't let me add as reviewers for some reason:
@faizan-m @zeroomega

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,8 @@
 FROM python:3.9
 
 COPY . /ml-compiler-opt
-RUN pip install --no-cache-dir -r /ml-compiler-opt/requirements.txt
+WORKDIR /ml-compiler-opt
+RUN pip install pipenv && pipenv sync --system && pipenv --clear
 
 WORKDIR /ml-compiler-opt/compiler_opt/tools
 ENV PYTHONPATH=/ml-compiler-opt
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,7 +1,8 @@
 FROM python:3.9
 
 COPY . /ml-compiler-opt
-RUN pip install --no-cache-dir -r /ml-compiler-opt/requirements.txt
+WORKDIR /ml-compiler-opt
```
petrhosek, line 4: I'd move the `WORKDIR` line above the `COPY` and use the `COPY . .` pattern, that way we're not duplicating the `/ml-compiler-opt` path which is a common pattern:

```
WORKDIR /ml-compiler-opt
COPY . .
```

petrhosek (changes requested):

Dockerfile
boomanaiden154: Fixed. Thanks for the pointer.

petrhosek (approved):

PR #197, opened 2023-01-04
Make TensorValue pickle-able

mtrofin:
Logs from optimizations like regalloc need to be "chunked" per context (i.e. per function, in regalloc's case). The bulk of the new log format is the tensor buffers. We could factor the log reader to allow parsing of context sections, but it's simpler (implementation-wise) to convert the whole log on the compilation runner side and then pass python objects around - since the bulk of the data stays unchanged (the bytes buffer underlying the tensor values), and, thus, the overhead is minimal.


mtrofin:
This looks like the wrong PR

PR #199, opened 2023-01-10
Added `spec` back to TensorValue

mtrofin:
This is because we will need the name to construct `Trajectory`es, and, while we could fetch it from elsewhere (like from the header), it's just easier to have it attached to the object, and the cost is small. In particular, in lists of repeated `TensorValue`s, pickling ensures only one copy of an object is made.


ebrevdo:
LGTM.

PR #200, opened 2023-01-10
Further parameterize data_reader_test

mtrofin:
Make the data source a parameter. Also remove `_agent_name` from the test state, it's really something specific to each state (and, thus, not needed as a field)

### DIFF
--- a/compiler_opt/rl/data_reader_test.py
+++ b/compiler_opt/rl/data_reader_test.py
@@ -66,7 +66,6 @@ def _write_tmp_tfrecord(file_path, example, num_examples):
 class DataReaderTest(tf.test.TestCase, parameterized.TestCase):
 
   def setUp(self):
-    self._agent_name = constant.AgentName.DQN
     observation_spec = {
         'feature_key':
             tf.TensorSpec(dtype=tf.int64, shape=(), name='feature_key')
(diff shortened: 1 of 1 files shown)


compiler_opt/rl/data_reader_test.py
```diff
@@ -88,23 +87,29 @@ def setUp(self):
...
 
-  @parameterized.named_parameters(
-      ('SequenceExampleDatasetFn',
-       data_reader.create_sequence_example_dataset_fn),
-      ('TFRecordDatasetFn', data_reader.create_tfrecord_dataset_fn))
-  def test_create_dataset_fn(self, test_fn):
-    example = _define_sequence_example(
-        self._agent_name, is_action_discrete=True)
+  def _create_sequence_example_datasource(self, example):
+    return [example.SerializeToString() for _ in range(100)]
```
ebrevdo, line 91: nit: [example.SerializeToString()] * 100  since you don't need to rerun identical function 100 times here.

compiler_opt/rl/data_reader_test.py
```diff
@@ -88,23 +87,29 @@ def setUp(self):
...
+
+  def _create_tfrecord_datasource(self, example):
```
ebrevdo, line 93: **Optional**: you can a temporary object that cleans up after itself if used in a "with" statement?

for file-based sources you can use https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile

for array based sources, you can just use return contextlib.nullcontext(array)

then use it via

with data_source_fn(...) as data_source:
   ... use data_source here

ebrevdo (approved):
just nits; LGTM so long as tests continue passing :)

compiler_opt/rl/data_reader_test.py
mtrofin: get_temp_dir() does that - the test infrastructure cleans it up after a test method is executed.

PR #201, opened 2023-01-11
Convert compiler logs to SequenceExamples

mtrofin:


compiler_opt/rl/log_reader_test.py
```diff
@@ -59,31 +62,34 @@ def create_example(fname: str):
...
-
-    f.write(json_to_bytes({'observation': 1}))
-    f.write(nl)
-    write_buff(f, t0_val, ctypes.c_float)
-    write_buff(f, t1_val, ctypes.c_int64)
-    f.write(nl)
-    f.write(json_to_bytes({'outcome': 1}))
-    f.write(nl)
-    write_buff(f, s, ctypes.c_float)
+    for ctx_id in range(nr_contexts):
```
ebrevdo, line 65: this looks like assembly to me.... you need to add comments for each line and/or wrap in functions with descriptions

mtrofin: yup - good point, fixed.

compiler_opt/rl/log_reader_test.py
```diff
@@ -158,6 +181,83 @@ def test_pickling(self):
...
+          value: 1.399999976158142
+          value: 1.5
+          value: 1.600000023841858
+        }
+      }
+    }
+  }
+}
+""", tf.train.SequenceExample())
+    self.assertEqual(expected_ctx_0, seq_examples['context_nr_0'])
```
ebrevdo, line 261: use `self.assertProtoEquals` which normalizes floating point numbers when comparing.

https://www.tensorflow.org/api_docs/python/tf/test/TestCase#assertProtoEquals

you will need to subclass from tf.test.TestCase and below use tf.test.main() instead of using absltest.

mtrofin: done

ebrevdo (approved):
writing json_to_bytes to the stream seems like a bit low-level for python (i'm not used to it)... but so long as round-trip tests pass LGTM.

my approach would have been to send a single msgpack containing a list of entries (one for each context) -- all at once.

mtrofin:
> writing json_to_bytes to the stream seems like a bit low-level for python (i'm not used to it)... but so long as round-trip tests pass LGTM.
> 
> my approach would have been to send a single msgpack containing a list of entries (one for each context) -- all at once.

One of the goals of the new format was to avoid the need to store anything on the compiler, for 2 reasons: efficiency is one; and the other is that we can reuse the exact same logic to communicate in "gym" setups.

PR #203, opened 2023-01-11
Enable using the new compiler log format

mtrofin:


compiler_opt/rl/inlining/inlining_runner.py
```diff
@@ -110,8 +110,10 @@ def compile_fn(
       if reward_only:
         return {_DEFAULT_IDENTIFIER: (None, native_size)}
 
-      with io.open(log_path, 'rb') as f:
-        sequence_example.ParseFromString(f.read())
+      result = log_reader.read_log_as_sequence_examples(log_path)
+      if len(result) != 1:
+        return {}
+      sequence_example = result.values()[0]
```
ebrevdo, line 116: suggest next(iter(result.values())).

if result is an actual dict, calling values()[0] would cause an error since values is an iterator.

ebrevdo (approved):

compiler_opt/rl/inlining/inlining_runner.py
mtrofin: ack - we need some shape of tests here.

mtrofin: well - pytype caught it, that's good!

PR #204, opened 2023-01-13
Drop pickling support for TensorValue, and reuse tf.TensorSpec

mtrofin:
Since we don't need to pass around TensorValues, as we go immediately to a Tensorflow object model, we don't need to worry about pickling. Also, `log_reader.TensorSpec` was too similar to `tf.TensorSpec` - might as well use the latter.

### DIFF
--- a/compiler_opt/rl/log_reader.py
+++ b/compiler_opt/rl/log_reader.py
@@ -65,17 +65,25 @@ from typing import Any, BinaryIO, Dict, Generator, List, Optional
 import sys
 import tensorflow as tf
 
-_element_types = {
-    'float': ctypes.c_float,
-    'double': ctypes.c_double,
-    'int8_t': ctypes.c_int8,
-    'uint8_t': ctypes.c_uint8,
-    'int16_t': ctypes.c_int16,
-    'uint16_t': ctypes.c_uint16,
-    'int32_t': ctypes.c_int32,
-    'uint32_t': ctypes.c_uint32,
-    'int64_t': ctypes.c_int64,
-    'uint64_t': ctypes.c_uint64
+_element_type_name_map = {
+    'float': (ctypes.c_float, tf.float32),
+    'double': (ctypes.c_double, tf.float64),
+    'int8_t': (ctypes.c_int8, tf.int8),
+    'uint8_t': (ctypes.c_uint8, tf.uint8),
+    'int16_t': (ctypes.c_int16, tf.int16),
+    'uint16_t': (ctypes.c_uint16, tf.uint16),
+    'int32_t': (ctypes.c_int32, tf.int32),
+    'uint32_t': (ctypes.c_uint32, tf.uint32),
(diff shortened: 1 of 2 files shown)


compiler_opt/rl/log_reader_test.py
```diff
@@ -159,29 +143,6 @@ def test_read_log(self):
       obs_id += 1
     self.assertEqual(obs_id, 2)
 
-  def test_pickling(self):
```
ebrevdo, line 162: do you need a unit test for serialization of tf.TensorSpec, or is that wrapped up in one of the other tests now?

compiler_opt/rl/log_reader.py
```diff
@@ -233,10 +222,10 @@ def read_log(fname: str) -> Generator[Record, None, None]:
...
 
 
-def _add_feature(se: tf.train.SequenceExample, spec: TensorSpec,
+def _add_feature(se: tf.train.SequenceExample, spec: tf.TensorSpec,
                  value: TensorValue):
   f = se.feature_lists.feature_list[spec.name].feature.add()
-  if spec.element_type in [ctypes.c_double, ctypes.c_float]:
+  if spec.dtype in [tf.float32, tf.float64]:
     lst = f.float_list.value
   else:
```
ebrevdo, line 230: raise an error if it's not an int format like int8, int32, int64?

mtrofin: done. it should never happen, but doesn't hurt either (defensive programming whatnot)

compiler_opt/rl/log_reader_test.py
mtrofin: no - we don't need to serialize anything anymore since we go straight to tf.train.SequenceExamples. Basically the object model in log_reader is internal.

Come to think of it, I'll make it internal.

mtrofin: made most APIs internal

yilei:
Thank you!

mtrofin:
is this good to go? thanks!

compiler_opt/rl/log_reader.py
```diff
@@ -165,38 +143,34 @@ def __getitem__(self, index: int):
...
 
 
 @dataclasses.dataclass(frozen=True)
-class Record:
+class ObservationRecord:
   context: str
   observation_id: int
-  feature_values: List[TensorValue]
-  score: Optional[TensorValue]
+  feature_values: List[_TensorValue]
```
ebrevdo, line 168: nit, up to you: ObservationRecord is returned to the user (externally visible) but types are currently _Hidden (_TensorValue).  consider moving it back to public.

ebrevdo (approved):

compiler_opt/rl/log_reader.py
mtrofin: renamed it to draw even more contrast to `tf.Tensor`

PR #207, opened 2023-01-18
Bump joblib to 1.1.1

mtrofin:
Rationale: https://security.snyk.io/vuln/SNYK-PYTHON-JOBLIB-3027033


petrhosek (approved):

boomanaiden154:
We should be able to delete this file as it doesn't get used in any of the normal install processes anymore (after the move to `pipenv`). We also need to bump this dependency within `Pipfile` and regenerate the transitive closure in `pipfile.lock`. Give me a couple minutes and I should have a PR ready.

PR #214, opened 2023-04-12
Fixes the swap of data and feature dimension to work in the general case.

kshiteejm:
Previous implementation was broken as using transpose assumes that `data_list` is a 2D array.

However, in certain cases (when all the feature values array lengths are the same) the `data_list` can be a 3D array as the call to `data_list = np.array(list(dataset.as_numpy_iterator()), dtype=object)`  merges inner np arrays and converts `data_list` into one big 3D array.

### DIFF
--- a/compiler_opt/tools/sparse_bucket_generator.py
+++ b/compiler_opt/tools/sparse_bucket_generator.py
@@ -170,7 +170,7 @@ def main(_) -> None:
   parser_fn = create_tfrecord_parser_fn(sequence_features)
   dataset = dataset.map(parser_fn, num_parallel_calls=tf.data.AUTOTUNE)
   data_list = np.array(list(dataset.as_numpy_iterator()), dtype=object)
-  data_list = np.transpose(data_list, [1, 0])
+  data_list = data_list.swapaxes(0, 1)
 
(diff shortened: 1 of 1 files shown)


mtrofin:
Thanks, @kshiteejm!

To capture offline discussion and @kshiteejm's offline example:

it's how numpy behaves:
```
>>> np.array(list([[np.ones(2)], [np.ones(2)], [np.ones(2)]]), dtype=object)
array([[[1.0, 1.0]],

       [[1.0, 1.0]],

       [[1.0, 1.0]]], dtype=object)

>>> np.array(list([[np.ones(2)], [np.ones(2)], [np.ones(3)]]), dtype=object)
array([[array([1., 1.])],
       [array([1., 1.])],
       [array([1., 1., 1.])]], dtype=object)
```

In our case, the data is shaped as: 
  * 1st dimension is traces (i.e. 1 per module)
  * 2nd dimension is features
  * 3rd dimension is feature tensor values

The general case is that the feature values have different shapes, in which case the data would be 2D with object values (the objects being various sized arrays). But if all features have exactly the same length, the result appears as a 3D value.

Thanks, @kshiteejm, for this clarification!

PR #219, opened 2023-05-01
Simplify scheduling work on a worker pool

mtrofin:
The user can just specify how to call a worker with some work, instead of having to go through the trouble of creating a work factory and so on.

### DIFF
--- a/compiler_opt/distributed/buffered_scheduler.py
+++ b/compiler_opt/distributed/buffered_scheduler.py
@@ -19,7 +19,7 @@
 import concurrent.futures
 import threading
 
-from typing import Any, Iterable, List, Callable, Tuple, TypeVar
+from typing import Any, Callable, Iterable, List, Optional, Tuple, TypeVar
 
 from compiler_opt.distributed import worker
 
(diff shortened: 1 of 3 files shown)


compiler_opt/distributed/buffered_scheduler.py
```diff
@@ -80,3 +81,32 @@ def chain_work(wkr: T):
       chain_work(w)
 
   return results
+
+
+def schedule_on_worker_pool(
+    action: Callable[[W, T],
+                     Any], jobs: Iterable[T], worker_pool: worker.WorkerPool
```
jacob-hegna, line 88: maybe pass through a `**kwargs` argument to `schedule` so that the user can set the `buffer` kwarg if they want (or just add a single `buffer` argument, but then if there are any more arguments added to `schedule` in the future we'd also have to hook them through here)

jacob-hegna (commented):
LGTM just one nit

compiler_opt/distributed/buffered_scheduler.py
mtrofin: done, exposing it explicitly.

Issue #217, opened 2023-05-01
Demo build instructions


reedkotler:
1. When I try and build the inlining demo, the instructions work until the need for "cipd". This is not a command I have
2. In the prelliminaries section, there is a typo: export WORKINGD_DIR=~ . It should be export WORKING_DIR=~

reedkotler:
In regalloc demo, build instructions failed. 
cd src
git apply $WORKING_DIR/ml-compiler-opt/experimental/chromium-bitcode-embedding.patch

error:
git apply $WORKING_DIR/ml-compiler-opt/experimental/chromium-bitcode-embedding.patch
error: patch failed: build/config/clang/clang.gni:15
error: build/config/clang/clang.gni: patch does not apply

mtrofin:
IIRC `cipd` comes from [here](https://github.com/google/ml-compiler-opt/blob/main/docs/inlining-demo/demo.md), specifically the link to get their source. At a point one of the packages there will add `cipd`. Attn @petrhosek in case that has changed on the Fuchsia side.

Re. the chromium patch - @boomanaiden154 , ptal? IIRC you updated the patches recently, but maybe we can specify in the demo what chromium git hash they are guaranteed to apply at?

boomanaiden154:
The `cipd` package should come from the Fuchsia setup instructions (https://fuchsia.dev/fuchsia-src/get-started/get_fuchsia_source), specifically the part when `.jiri_root` is added to the path. I'll take a deeper look soon and probably add a note since I know other people have had trouble with it in the past. I'll try and push rebased Chromium patches too along with specifying a Chromium commit SHA (maybe working on upstreaming too).

boomanaiden154:
Closing this issue since this should be resolved with the patches that landed recently. Let us know if you have any more issues!

PR #220, opened 2023-05-02
Update docs to point out common issue with Fuchsia path setup

boomanaiden154:
Should fix the rest of issue #217.

### DIFF
--- a/docs/inlining-demo/demo.md
+++ b/docs/inlining-demo/demo.md
@@ -71,6 +71,14 @@ cipd install fuchsia/sysroot/linux-arm64 latest -root ${SYSROOT_DIR}/linux-arm64
 cipd install fuchsia/sysroot/linux-amd64 latest -root ${SYSROOT_DIR}/linux-x64
 ```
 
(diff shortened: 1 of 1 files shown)


docs/inlining-demo/demo.md
```diff
@@ -71,6 +71,14 @@ cipd install fuchsia/sysroot/linux-arm64 latest -root ${SYSROOT_DIR}/linux-arm64
 cipd install fuchsia/sysroot/linux-amd64 latest -root ${SYSROOT_DIR}/linux-x64
 ```
 
+**Note:** If your shell can't find the `cipd` command, it's likely your `$PATH`
+variable doesn't contain the path to `.jiri_root/bin`. To add this to the path,
+run the following command:
+
+```shell
+export PATH=$PATH:$(realpath ~/fuchsia/.jiri_root/bin)
```
petrhosek, line 79: This should be `$FUCHSIA_SRCDIR/.jiri_root/bin` to be consistent with the rest of the document.

PR #222, opened 2023-05-02
Switch `generate_default_trace` to `worker` infrastructure.

mtrofin:


compiler_opt/tools/generate_default_trace.py
```diff
@@ -160,74 +145,60 @@ def main(_):
...
-            performance_writer.write(
-                (f'{module_name},{key},{value.default_reward},'
-                 f'{value.moving_average_reward}\n'))
-
-      print((f'{total_successful_examples} of {len(corpus_elements)} modules '
-             f'succeeded, and {total_training_examples} trainining examples '
-             'written'))
-      for p in processes:
-        p.join()
+      with local_worker_manager.LocalWorkerPoolManager(
```
jacob-hegna, line 160: can you abstract this away (similar to train_locally.py) so we can use larger worker pools for generating traces?

mtrofin: done

jacob-hegna (approved):
LGTM!

PR #228, opened 2023-05-04
Add MLGO environments.

jacob-hegna:


compiler_opt/rl/inlining/config.py
```diff
@@ -99,3 +103,54 @@ def observation_processing_layer(obs_spec):
 
 def get_nonnormalized_features():
   return ['reward', 'inlining_default', 'inlining_decision']
+
+
+class InliningForSizeTask(env.MLGOTask):
```
mtrofin, line 108: can you move this to its own file, like an inlining_task.py, config.py has just the ml-facing things.

compiler_opt/rl/env.py
```diff
@@ -0,0 +1,373 @@
...
+
+  def _reward_fn(a: float, b: float) -> float:
+    return 1.0 - (a + _EPS) / (b + _EPS)
+
+  assert score_a.keys() == score_b.keys()
+  return {key: _reward_fn(score_a[key], score_b[key]) for key in score_a}
+
+
+@contextlib.contextmanager
+def clang_session(
```
mtrofin, line 220: make this gin configurable, and the MLGOTasks, too. Then the task_factory can just be a `type`. 

You could make clang_path be part of the ctor of the MLGOTask (basically following the CompilationRunner way for this)

I'd also leave command line construction to the MLGOTask implementation, so maybe just ask the task object: here's the tempdir, here's your baseline cmdline (i.e. the one coming from LoadedModuleSpec), and here's the interactive pipe name (None if you don't want that) or the policy path (None if you want interactive) - maybe best to wrap these 2 into a little class enforcing that _both_ is bad.

compiler_opt/rl/env.py
```diff
@@ -0,0 +1,373 @@
...
+
+    compiled_module_path = os.path.join(td, _COMPILED_MODULE_NAME)
+    cmdline = ([clang_path] + list(module.build_command_line(td)) +
+               task.get_clang_args() + ['-o', compiled_module_path])
+    if interactive:
+      cmdline += task.get_interactive_clang_args(
+          os.path.join(td, _INTERACTIVE_PIPE_FILE_BASE))
+
+    def _get_scores() -> dict[str, float]:
+      assert os.path.exists(compiled_module_path)
```
mtrofin, line 254: I'd let the MLGOTask figure out where to get the scores from.

jacob-hegna: they can get the scores from wherever they want, this is just making sure that we actually finishing compiling the module before calling the method

mtrofin (approved):

PR #231, opened 2023-05-09
Explicitely set library directory name

MatzeB:
When running `cmake` on Fedora/CentOS the default library directory is "lib64" instead of "lib" as seen on ubuntu/debian. Force the name "lib" to match expectations of later parts in the `build_tflite.sh` script.


mtrofin:
lgtm, @petrhosek any issues?

PR #233, opened 2023-05-09
Write reward_stat_map atomically

MatzeB:
Write reward_stat_map to separate file and atomically rename to avoid data corruption when backing up an instance/snapshots while a new version of the file is written.


mtrofin:
I'm surprised there's a race condition here, where is it coming from?

MatzeB:
I was running training on cloud machines that could get killed/interrupted at any point in time. So I just `rsync`ed the current model state away every couple minutes hoping to continue from the last checkpoint if the instance was killed early.

This scheme failed on me twice as it happened to copy a half-written reward_stat_map file...

mtrofin:
Aah! makes sense.

PR #235, opened 2023-05-09
compilation_runner: Option to keep temporary files

MatzeB:
This allows users to manually re-run commands after the fact to investigate failures.

### DIFF
--- a/compiler_opt/rl/inlining/inlining_runner.py
+++ b/compiler_opt/rl/inlining/inlining_runner.py
@@ -47,7 +47,8 @@ class InliningRunner(compilation_runner.CompilationRunner):
 
   def compile_fn(
       self, command_line: corpus.FullyQualifiedCmdLine, tf_policy_path: str,
-      reward_only: bool) -> Dict[str, Tuple[tf.train.SequenceExample, float]]:
+      reward_only: bool,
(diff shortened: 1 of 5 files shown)


compiler_opt/rl/inlining/inlining_runner.py
```diff
@@ -68,56 +69,52 @@ def compile_fn(
...
     """
 
-    working_dir = tempfile.mkdtemp()
+    working_dir = tempfile.mkdtemp(dir=workdir)
 
     log_path = os.path.join(working_dir, 'log')
     output_native_path = os.path.join(working_dir, 'native')
 
     native_size = 0
-    try:
```
mtrofin, line 77: why don't we need the try/catch?

mtrofin: got it. 

Can you also change the `regalloc_priority_runner.py`?

MatzeB: I can do a best effort (since the thing is not in a runnable state right now as far as I can tell).

mtrofin: Thanks - mostly about the second inner dir creation, for consistency.

PR #237, opened 2023-05-09
Refactor agent configuration.

mtrofin:
Replace agent ID and if-checks in various parts of the code (in agent creation and in data processing, respectivelly) with `AgentConfig`, an abstraction encapsulating the various behaviors.


jacob-hegna (approved):
Didn't read every line very closely because it's purely code movement; just looked at the major themes of the refactor/made sure all the places I remembered mattering were changed. Assuming tests pass LGTM

PR #242, opened 2023-05-16
[NFC] Log reader changes for MLGO environments.

jacob-hegna:


compiler_opt/rl/log_reader.py
```diff
@@ -86,6 +86,11 @@
 }
 
 
+def convert_dtype_to_ctype(dtype: str) -> Any:
```
mtrofin, line 89: `-> Tuple(type, tf.DType)` or something

compiler_opt/rl/log_reader.py
```diff
@@ -74,7 +74,7 @@
     'int32_t': (ctypes.c_int32, tf.int32),
     'uint32_t': (ctypes.c_uint32, tf.uint32),
     'int64_t': (ctypes.c_int64, tf.int64),
-    'uint64_t': (ctypes.c_uint64, tf.uint64)
+    'uint64_t': (ctypes.c_uint64, tf.uint64),
```
mtrofin, line 77: is the comma necessary?

compiler_opt/rl/log_reader.py
```diff
@@ -120,6 +127,14 @@ def __init__(self, spec: tf.TensorSpec, buffer: bytes):
   def spec(self):
     return self._spec
 
+  @property
+  def buffer(self):
```
mtrofin, line 131: why do you need this and the `len` property? should they have a small unit test, too? 

also, maybe `buffer` -> `raw_bytes` to make it clear it's that, not another way to dereference the typed view

and then `len` -> do you need to use it as a length of the raw buffer? if so, maybe call it like that and return the _len multiplied by the scalar type size.

compiler_opt/rl/log_reader.py
jacob-hegna: fixed

compiler_opt/rl/log_reader.py
```diff
@@ -95,7 +100,8 @@ def create_tensorspec(d: Dict[str, Any]) -> tf.TensorSpec:
   return tf.TensorSpec(
       name=name,
       shape=tf.TensorShape(shape),
-      dtype=_element_type_name_to_dtype[element_type_str])
+      dtype=_element_type_name_to_dtype[element_type_str],
```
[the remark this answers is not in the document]
jacob-hegna: fixed

compiler_opt/rl/log_reader.py
jacob-hegna: I convert the data given by the log_reader to a numpy array, and the easiest way to do that was to use https://numpy.org/doc/stable/reference/generated/numpy.frombuffer.html.

I use numpy in the environments instead of tensorflow because numpy is universally used in every ML framework (tf, pytorch, jax) and I want to start decoupling things from TF when they don't fundamentally require it.

I'm not sure if they need unit tests - if you want I can add them, but they're not doing anything surprising. I think if the other tests fail it should catch any issues here.

compiler_opt/rl/log_reader.py
```diff
@@ -229,7 +252,7 @@ def _add_feature(se: tf.train.SequenceExample, spec: tf.TensorSpec,
 
 
 def read_log_as_sequence_examples(
-    fname: str) -> Dict[str, tf.train.SequenceExample]:
+    fname: str,) -> Dict[str, tf.train.SequenceExample]:
```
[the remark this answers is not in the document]
jacob-hegna: fixed

jacob-hegna (commented):
The reason for all the spurious changes is that I realized my vim was autoformatting the python files not using yapf, but instead using the internal google formatting tool... so I would save the file, thinking it was formatting, upload the commit, see that yapf failed in CI, then use yapf and reupload. So, all the spurious changes were because of the google internal formatter and yapf (also a google python formatter) fighting each other. went back through and tried to revert all the spurious things.

compiler_opt/rl/log_reader.py
mtrofin: But the object itself acts as a buffer, it has an indexer and a __len__.

jacob-hegna: np has its own notion of a buffer interface which is pretty specific, and if we naively try to pass one of these objects to `np.frombuffer` we get the error:
```
 TypeError: a bytes-like object is required, not 'LogReaderTensorValue'
```

mtrofin: ha! interesting. OK. so can we do with just a `raw_bytes`, looks like len() wouldn't be needed?

jacob-hegna: ok, replaced `raw_bytes` and `len` with a `to_numpy` method directly, and added a unit test for it.

mtrofin: a, neat. Thanks!

mtrofin:
lgtm

Issue #248, opened 2023-06-02
Plan to port Evolutionary Strategy algorithm


salaast:
This is a port of the work by: Krzysztof Choromanski, Mark Rowland, Vikas Sindhwani, Richard E. Turner, Adrian Weller:  "Structured Evolution with Compact Architectures for Scalable Policy Optimization", https://arxiv.org/abs/1804.02395

This plan will follow 3 main steps:
1. Port existing codebase
2. Add type annotations, and idioms along ml-compiler-opt codebase
3. Create training pipeline for Evolutionary Strategy

PR #249, opened 2023-06-02
Add blackbox optimizers and secondorder blackbox optimizers

salaast:
This is a port of the work by: Krzysztof Choromanski, Mark Rowland,
Vikas Sindhwani, Richard E. Turner, Adrian Weller:  "Structured Evolution
with Compact Architectures for Scalable Policy Optimization",
https://arxiv.org/abs/1804.02395

First step in Issue #248

### DIFF
--- a/compiler_opt/es/combined_blackbox_optimizers.py
+++ b/compiler_opt/es/combined_blackbox_optimizers.py
@@ -1,0 +1,1108 @@
+# coding=utf-8
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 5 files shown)


compiler_opt/es/combined_blackbox_optimizers.py
mtrofin, line 1: I'd rename the file / package to `blackbox_optimizers`, the fact that they happen to be "combined" (as in, the 3 we need all in 1 package) is not that interesting.

mtrofin (commented):
you can probably remove the "depends on pr #247 " from the message

mtrofin (approved):

PR #259, opened 2023-06-09
Add SamplerWithoutReplacement to corpus.py.

jacob-hegna:
This is useful for supervised learning contexts where we want to train in fixed epochs, where in each epoch we visit each module in the corpus exactly once.

### DIFF
--- a/compiler_opt/rl/local_data_collector_test.py
+++ b/compiler_opt/rl/local_data_collector_test.py
@@ -23,12 +23,13 @@ from typing import List, Tuple
 import tensorflow as tf
 # This is https://github.com/google/pytype/issues/764
 from google.protobuf import text_format  # pytype: disable=pyi-error
-from tf_agents.system import system_multiprocessing as multiprocessing
+from compiler_opt.distributed.local.local_worker_manager import LocalWorkerPoolManager
+from compiler_opt.rl import compilation_runner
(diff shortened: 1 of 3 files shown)


compiler_opt/rl/local_data_collector_test.py
```diff
@@ -21,16 +21,14 @@
...
-from compiler_opt.rl import compilation_runner
-from compiler_opt.rl import corpus
-from compiler_opt.rl import data_collector
-from compiler_opt.rl import local_data_collector
-from compiler_opt.rl import policy_saver
+from tf_agents.system import system_multiprocessing as multiprocessing
+
+from compiler_opt.distributed.local.local_worker_manager import \
+    LocalWorkerPoolManager
+from compiler_opt.rl import (compilation_runner, corpus, data_collector,
```
mtrofin, line 30: I think we want these one per line, for some stylistic reason that I now realize isn't supported by pylint of yapf (or we didn't figure out how to express it)

mtrofin (approved):

compiler_opt/rl/local_data_collector_test.py
jacob-hegna: oh sorry, this is another discrepancy between my vim on-save autoformatter and yapf. will fix

PR #260, opened 2023-06-09
Fix extract_ir.py for macos.

jacob-hegna:
For whatever reason, the section names for .llvmcmd and .llvmbc are different on macos. One problem is that macos uses macho object files instead of elf, so in addition to the section names there are segment names that need to be specified. The other problem is that regardless of the segments, the section names are different. This patch makes the segment name configurable by a flag, and gives sensible defaults for linux + documentation for macos.

### DIFF
--- a/compiler_opt/tools/extract_ir.py
+++ b/compiler_opt/tools/extract_ir.py
@@ -78,15 +78,15 @@ flags.mark_flags_as_required(['output_dir'])
 
 flags.DEFINE_string(
     'cmd_section_name', '.llvmcmd',
-    'Name of the section name passed to llvm-objcopy. For linux systems, the '
-    'default .llvmcmd is correct. On macos, one should use __LLVM,__cmdline'
-)
+    'The section name passed to llvm-objcopy. For ELF object files, the '
+    'default .llvmcmd is correct. For Mach-O object files, one should use '
+    'something like __LLVM,__cmdline')
 
 flags.DEFINE_string(
     'bitcode_section_name', '.llvmbc',
-    'Name of the section name passed to llvm-objcopy. For linux systems, the '
-    'default .llvmbc is correct. On macos, one should use __LLVM,__bitcode'
-)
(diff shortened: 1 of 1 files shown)


mtrofin (approved):

compiler_opt/tools/extract_ir.py
```diff
@@ -371,4 +380,5 @@ def main(argv):
 
 
 if __name__ == '__main__':
+  multiprocessing.set_start_method('fork')
```
mtrofin, line 383: why?

compiler_opt/tools/extract_ir.py
```diff
@@ -76,6 +76,18 @@
     'passed in the local case.')
 flags.mark_flags_as_required(['output_dir'])
 
+flags.DEFINE_string(
+    'cmd_section_name', '.llvmcmd',
+    'Name of the section name passed to llvm-objcopy. For linux systems, the '
+    'default .llvmcmd is correct. On macos, one should use __LLVM,__cmdline'
```
petrhosek, line 82: Technically this is not about the OS but about the file format (for example, you might be running training on macOS but using binaries compiled for Linux), so I think it's more accurate to say "ELF" and "Mach-O" rather than "Linux" and "macOS" respectively.

compiler_opt/tools/extract_ir.py
jacob-hegna: oh, meant to include this in the pull request description. it's because of this
https://github.com/abseil/abseil-py/issues/170

basically, the default method of multiprocessing on macos is spawn instead of fork, which means the flags don't get passed around correctly. The flags are accessed within the spawned processes, so without this line macos will use default flag values for anything inside the subprocess.

compiler_opt/tools/extract_ir.py
jacob-hegna: fixed - thanks!

PR #262, opened 2023-06-09
Fix caching in CI

boomanaiden154:
This fixes the caching for python dependencies in the CI by making pipenv install in a virtual environment and telling the setup-python action to cache this venv as it doesn't currently support the global pipenv cache. This increases the speed of some of the fast jobs (like the yapf format checker) quite a bit. Formatting went from about 2.5min to about 0.7min in my testing. This assumes that the cache is warm.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -52,15 +52,17 @@ jobs:
         uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python-version }}
-          cache: 'pip'
+          cache: 'pipenv'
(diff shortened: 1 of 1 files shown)


boomanaiden154:
Once this gets merged I'll go and purge the current caches because they end up being the same since `Pipfile.lock` hasn't changed, so it thinks it has a cache hit, but there isn't anything in the cache currently until this commit gets merged into main.

The rabbit holes that the "one line cleanup that definitely shouldn't break anything patch" will lead you down.

PR #263, opened 2023-06-13
Add pyproject.toml to enable packaging

boomanaiden154:
Adds a basic `pyproject.toml` so that we can package the ML python tooling for use outside of the repository.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,22 @@
+[build-system]
+requires = ["setuptools"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "ml-compiler-opt"
+description = "Tooling for ML in LLVM"
+readme = "README.md"
+requires-python = ">=3.8,<3.11"
+dependencies = [
+  "absl-py>=1.0.0",
+  "gin-config>=0.5.0",
+  "psutil>=5.9.0",
+  "tf-agents>=0.16.0",
+  "tensorflow>=2.12.0",
+  "dm-reverb>=0.11.0"
+]
+version="0.0.1"
+
+[tool.setuptools.packages.find]
+include = ["compiler_opt*"]
+
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,8 @@
 __pycache__
 .vscode
 .idea
+
+# Python packaging build directories
+build/*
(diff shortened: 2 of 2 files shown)


pyproject.toml
```diff
@@ -0,0 +1,22 @@
+[build-system]
+requires = ["setuptools"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "ml-compiler-opt"
+description = "Tooling for ML in LLVM"
+readme = "README.md"
+requires-python = ">=3.8"
```
mtrofin, line 9: can this say <=3.10 as well? Not relying much on py backcompat.

pyproject.toml
```diff
@@ -0,0 +1,22 @@
...
+dependencies = [
```
mtrofin, line 10: can these be versioned, or shared with the pipenv file?

pyproject.toml
boomanaiden154: Updated it to say `<3.11`. `<=3.10` doesn't allow for any of the `3.10.x` minor releases.

pyproject.toml
boomanaiden154: I've added minimum version information to the packages (not specifying exact versions to try and avoid breaking installations with other packages that might have odd version requirements). I would like to share it with the `Pipfile`, but it's not trivial since `setuptools` only supports dynamic dependencies from a `requirements.txt`. I'll look into at some point, but it seems fairly standard to duplicate the deps here (e.g., [here](https://github.com/tensorflow/tensorflow/blob/321e541cbda1b4e473d264b4fe9bdd3dc2dbe488/tensorflow/tools/pip_package/setup.py#L84) in Tensorflow).

boomanaiden154:
@mtrofin Do you have any other comments?

.Pipfile.swp
mtrofin, line 1: what's up with this file

mtrofin (approved):
probably want to remove `.Pipfile.swp`?

.Pipfile.swp
boomanaiden154: Looks like I had another vim window open when I ran `git commit`. Probably should add `*.swp` to my global ignore at some point.

PR #268, opened 2023-06-20
Refactor combine_training_corpus.py into script and library

boomanaiden154:
This patch refactors combine_training_corpus.py into a library file that can easily be imported in downstream projects (and other utilities). This also makes unit testing slightly easier (as no special accomodations have to be made for CLI flags). This patch adds in two unittests for combining training corpora as well.


mtrofin (approved):
Thanks for also adding test!

I think we need to do a bit more work to actually support combining corpora with mixed thinlto-ness? Anyway, out of the scope of this patch.

PR #269, opened 2023-06-20
Add directory input to extract_ir.py

boomanaiden154:
This patch adds in a directory input to extract_ir.py where the script does a glob for *.o files inside the specified directory. This is useful in cases where it is easy to specify specific flags through a build system (like the necessary -Xclang -fembed-bitcode=all), but the build system produces no manifest of compilation commands like CMake's compile_commands.json. This isn't reccomended to use in place of a proper manifest but is necessary for supporting build systems like Make.

### DIFF
--- a/compiler_opt/tools/extract_ir.py
+++ b/compiler_opt/tools/extract_ir.py
@@ -41,10 +41,12 @@ from compiler_opt.tools import extract_ir_lib
 
 flags.DEFINE_string(
     'input', None,
-    'Input file - either compile_commands.json or a linker parameter list')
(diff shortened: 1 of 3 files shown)


compiler_opt/tools/extract_ir.py
```diff
@@ -41,10 +41,12 @@
 
 flags.DEFINE_string(
     'input', None,
-    'Input file - either compile_commands.json or a linker parameter list')
+    'Input file - either compile_commands.json, a linker parameter list, or a'
```
mtrofin, line 44: nit: "Input file, or input directory "

mtrofin (approved):

PR #270, opened 2023-06-20
Add SklearnRegressionBlackboxOptimizer class

salaast:


compiler_opt/es/blackbox_optimizers.py
```diff
@@ -324,34 +471,6 @@ def set_state(self, state: SequenceOfFloats) -> None:
 
 See class descriptions for more detailed notes on each algorithm.
 """
-
-_GRAD_TYPE = flags.DEFINE_string('grad_type', 'MC', 'Gradient estimator.')
```
mtrofin, line 328: can you do this removal in a separate PR? thanks.

mtrofin:
please also add the corresponding unit tests.

compiler_opt/es/blackbox_optimizers.py
```diff
@@ -263,7 +277,141 @@ def run_step(self, perturbations: npt.NDArray[np.float32],
...
+      current_state = [self.nb_steps]
+      current_state += self.sum_state_vector
+      current_state += self.squares_state_vector
+      current_state += self.mean_state_vector + self.std_state_vector
+      return current_state + gradient_ascent_state
+    else:
+      return gradient_ascent_state
+
+  def update_state(self, evaluation_stats: SequenceOfFloats) -> None:
+    if self.hyperparameters_update_method == UpdateMethod.STATE_NORMALIZATION:
```
mtrofin, line 298: and if not, then what?
i.e. is it a bug to call `update_state` if self.hyperparameters_update_method != UpdateMethod.STATE_NORMALIZATION, or the API is meant to be callable in all cases, for uniformity, and only does something in this case?

if the latter, maybe easier to read if:
```
if self.<blah blah> != UpdateMethod.STATE_NORMALIZATION:
  return

<current code>
```

less indenting :)

compiler_opt/es/blackbox_optimizers.py
```diff
@@ -263,7 +277,141 @@ def run_step(self, perturbations: npt.NDArray[np.float32],
...
+      mean_squares_state_vector = [
+          x / float(self.nb_steps) for x in self.squares_state_vector
+      ]
+
+      self.std_state_vector = [
+          math.sqrt(max(a - b * b, 0.0))
+          for a, b in zip(mean_squares_state_vector, self.mean_state_vector)
+      ]
+
+  def set_state(self, state: SequenceOfFloats) -> None:
```
mtrofin, line 321: ah, so these APIs on MonteCarlo weren't used before, correct?

do they have test?

compiler_opt/es/blackbox_optimizers.py
salaast: The latter is true so I will update the code to this style.

compiler_opt/es/blackbox_optimizers.py
salaast: They were here before and are present in the current codebase, but oddly the diff misrepresents it as new even after rebasing.

compiler_opt/es/blackbox_optimizers.py
salaast: The removal has been separated into a different PR now

compiler_opt/es/blackbox_optimizers.py
```diff
@@ -212,44 +226,112 @@ def set_state(self, state: SequenceOfFloats) -> None:
     raise NotImplementedError('Abstract method')
 
 
-class MonteCarloBlackboxOptimizer(BlackboxOptimizer):
-  """Class implementing GD optimizer with MC estimation of the gradient."""
+class CommonStateHandler(BlackboxOptimizer):
```
mtrofin, line 229: How about `StatefulOptimizer` or `AbstractStatefulOptimizer` - while it _is_ a common something or other, so would all intermediate base classes be - meaning, calling it that doesn't offer any information that's not implicit.

I'm proposing `StatefulOptimizer` because it seems the abstraction brings in state management?

(I realize "state" in CommonStateHandler may actually refer to the algorithm state, but the name also makes it sound so much like "this is a grabbag of some common fields")

compiler_opt/es/blackbox_optimizers.py
```diff
@@ -212,44 +226,112 @@ def set_state(self, state: SequenceOfFloats) -> None:
...
+  """Intermediate parent class for blackbox optimizers.
```
mtrofin, line 230: for _stateful_ optimizers?

mtrofin (approved):
lgtm, just the one naming nit

PR #273, opened 2023-06-22
Add docstrings to extract_ir_lib.py

boomanaiden154:
Now that extract_ir_lib.py is being used in other projects through the Pip package rather than just the ml-compiler-opt CLI tooling, it would be good to have better documentation on how to use these functions. This commit adds in doc strings for some of the functions that are most likely to be used in downstream projects and fixes up some other docstrings.

This was motivated by not passing an empty string instead of None to cmd_filter which led to some annoying debugging when other issues were at play.

### DIFF
--- a/compiler_opt/tools/extract_ir_lib.py
+++ b/compiler_opt/tools/extract_ir_lib.py
@@ -105,7 +105,9 @@ class TrainingIRExtractor:
 
   def _get_extraction_cmd_command(self, llvm_objcopy_path: str,
                                   cmd_section_name: str):
-    """Call llvm_objcopy to extract the llvmcmd section in self._cmd_file."""
+    """Get llvm-objcopy and process args to a produce a command string that,
+    when invoked, will extract the cmd section info ths self.cmd_file() file.
+    """
     return [
         llvm_objcopy_path,
         '--dump-section=' + cmd_section_name + '=' + self.cmd_file(),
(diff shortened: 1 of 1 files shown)


compiler_opt/tools/extract_ir_lib.py
```diff
@@ -105,7 +105,9 @@ def thinlto_index_file(self):
 
   def _get_extraction_cmd_command(self, llvm_objcopy_path: str,
                                   cmd_section_name: str):
-    """Call llvm_objcopy to extract the llvmcmd section in self._cmd_file."""
+    """Gets command to make llvm_objcopy extract the llvmcmd section in
+    self._cmd_file.
```
mtrofin, line 109: took me a bit to understand this, how about "Get llvm-objcopy and process args that, when invoked, would extract the llvmcmd section into the self.cmd_file() file"

or something like that. I was missing what flows where :)

compiler_opt/tools/extract_ir_lib.py
```diff
@@ -114,7 +116,9 @@ def _get_extraction_cmd_command(self, llvm_objcopy_path: str,
 
   def _get_extraction_bc_command(self, llvm_objcopy_path: str,
                                  bitcode_section_name: str):
-    """Call llvm_objcopy to extract the llvmbc section in self._bc_file."""
+    """Gets command to make llvm_objcopy extract the llvmbc section in
```
mtrofin, line 119: same as above

mtrofin (approved):

PR #276, opened 2023-06-24
Add ability to extract raw bitcode

boomanaiden154:
This patch adds in the ability to extract raw bitcode from a directory. This is motivated primarily by extracting LLVM IR bitcode from rust projects as it is quite easy to emit LLVM bitcode for each target but bitcode embedded in object files is much less trivial and neither solution currently provides command line arguments so they need to be reconstructed regardless.

This could probably benefit from some refactoring in the future to make it more clean. I held off on any additional refactoring before this patch as some of it was less trivial and I need some feedback at how specific IR extraction techniques are used. I believe we could unify the `--thinlto_build` and `--input_type` flags (assuming that distributed ThinLTO is always done with a linker parameter file which I'm unsure of) which might make things a little bit more clean. Some later renaming (i.e., renaming `obj_file()` to something more generic or using a different property might also make things more clean. If these things are desired before landing this patch I can make that work too.

### DIFF
--- a/compiler_opt/tools/make_corpus.py
+++ b/compiler_opt/tools/make_corpus.py
@@ -19,8 +19,8 @@ the following command:
 
 PYTHONPATH=$PYTHONPATH:. python3 ./compiler_opt/tools/make_corpus.py \
   --input_dir=<path to input directory> \
-  --output_dir=<path to output direcotry> \
-  --default_flags="<list of space separated flags>"
+  --output_dir=<path to output directory> \
+  --default_args="<list of space separated flags>"
 """
 
 from absl import app
@@ -32,7 +32,7 @@ from compiler_opt.tools import make_corpus_lib
 flags.DEFINE_string('input_dir', None, 'The input directory.')
 flags.DEFINE_string('output_dir', None, 'The output directory.')
 flags.DEFINE_string(
-    'default_flags', '',
+    'default_args', '',
     'The compiler flags to compile with when using downstream tooling.')
 
 flags.mark_flag_as_required('input_dir')
@@ -50,7 +50,7 @@ def main(_):
   make_corpus_lib.copy_bitcode(relative_paths, FLAGS.input_dir,
                                FLAGS.output_dir)
   make_corpus_lib.write_corpus_manifest(relative_paths, FLAGS.output_dir,
-                                        FLAGS.default_flags.split())
+                                        FLAGS.default_args.split())
 
 
 if __name__ == '__main__':
--- a/compiler_opt/tools/make_corpus.py
+++ b/compiler_opt/tools/make_corpus.py
@@ -1,0 +1,57 @@
+# coding=utf-8
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


compiler_opt/tools/extract_ir_lib.py
```diff
@@ -314,14 +350,20 @@ def run_extraction(objs: List[TrainingIRExtractor], num_workers: int,
...
   """
   extract_artifacts = functools.partial(
       TrainingIRExtractor.extract,
       llvm_objcopy_path=llvm_objcopy_path,
       cmd_filter=cmd_filter,
       thinlto_build=thinlto_build,
       cmd_section_name=cmd_section_name,
-      bitcode_section_name=bitcode_section_name)
+      bitcode_section_name=bitcode_section_name,
+      default_command_line_flags=default_command_line_flags,
```
mtrofin, line 365: can this notion of "default_command_line_flags" be joined with `global_command_override` below?

mtrofin (commented):
at this point, why not let this be a different tool that just makes the corpus_descriptor json file? it's not _extracting_ anything anymore, right?

boomanaiden154:
Yes, that's the plan. I just haven't had an opportunity to rewrite the patch. I'll have something up soon.

compiler_opt/tools/make_corpus.py
```diff
@@ -0,0 +1,57 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Tool for making a corpus from arbitrary bitcode.
+
+To create a corpus from a set of bitcode files in an input directory, run
+the following command:
+
+PYTHONPATH=$PYTHONPATH:. python3 ./compiler_opt/tools/make_corpus.py \
+  --input_dir=<path to input directory> \
+  --output_dir=<path to output direcotry> \
```
mtrofin, line 22: spelling

compiler_opt/tools/make_corpus_lib.py
```diff
@@ -0,0 +1,83 @@
...
+
+import pathlib
+import os
+import shutil
+import json
+
+from typing import List, Optional
+
+
+def load_bitcode_from_directory(bitcode_base_dir: str):
```
mtrofin, line 25: can add type annotation for return, e.g. Iterable[str] or what it is.

mtrofin: (same elsewhere)

compiler_opt/tools/make_corpus_lib.py
```diff
@@ -0,0 +1,83 @@
...
+  for relative_path in relative_paths:
+    base_path = os.path.join(bitcode_base_dir, relative_path)
+    destination_path = os.path.join(output_dir, relative_path)
+    os.makedirs(os.path.dirname(destination_path), exist_ok=True)
+    shutil.copy(base_path, destination_path)
+
+
+def write_corpus_manifest(relative_output_paths: List[str],
+                          output_dir: str,
+                          default_flags: Optional[List[str]] = None):
```
mtrofin, line 62: default_args? if more consistent with elsewhere?

mtrofin (approved):

PR #277, opened 2023-07-12
Fix spurious AgentConfig member reference

mtrofin:
This was post PR #237, unfortunately pytype doesn't detect these types of errors, and we don't have unit tests for these paths.


salaast (approved):
lgtm

ebrevdo (approved):

PR #278, opened 2023-07-14
Add logging verbosity level flag to extract_ir.py

boomanaiden154:
This commit adds a logging verbosity level flag to extract_ir.py and the appropriate plumbing/implementation in extract_ir_lib.py. This is primarily motivated by these errors coming up quite often in non-trivial builds (as it is fairly often there is some assembly or some flags don't get passed around somewhere) and not providing a very high signal to noise ratio, especially when used as a library against a bunch of projects at once.

I've implemented this as a custom flag rather than using absl's `logging.get_verbosity` and `logging.set_verbosity` as I think this approach is a bit more expressive for the purposes here, and it also avoids us needing to set global state which is nice within the project and especially for downstream users (although it's not a big switch to do so the other method). I'm open to switching over if others have different ideas.

### DIFF
--- a/compiler_opt/tools/extract_ir.py
+++ b/compiler_opt/tools/extract_ir.py
@@ -30,9 +30,9 @@ In a local ThinLTO case, the compilation is assumedto have been performed
 specifying -Wl,--save-temps=import -Wl,--thinlto-emit-index-files
 
 To change the logging verbosity, pass an integer representing the desired
-verbosity to the --verbosity flag. Use 0 for all logs and detailed debug
-information, -1 for solely warnings, and -2 to only print a status message at
-the end.
+verbosity to the --verbosity flag. Use 0 for all logs, status information, 
+and detailed debug information, -1 for solely warnings, and -2 to not produce
+any output.
 """
 
 import json
@@ -132,9 +132,6 @@ def main(argv):
 
   extract_ir_lib.write_corpus_manifest(FLAGS.thinlto_build,
                                        relative_output_paths, FLAGS.output_dir)
-
-  # Reset logging verbosity so that we print status messages
-  logging.set_verbosity(logging.INFO)
 
   logging.info('Converted %d files out of %d',
                len(objs) - relative_output_paths.count(None), len(objs))
--- a/compiler_opt/tools/extract_ir.py
+++ b/compiler_opt/tools/extract_ir.py
@@ -30,7 +30,7 @@ In a local ThinLTO case, the compilation is assumedto have been performed
 specifying -Wl,--save-temps=import -Wl,--thinlto-emit-index-files
 
 To change the logging verbosity, pass an integer representing the desired
-verbosity to the --verbosity flag. Use 0 for all logs, status information, 
+verbosity to the --verbosity flag. Use 0 for all logs, status information,
 and detailed debug information, -1 for solely warnings, and -2 to not produce
 any output.
 """
--- a/compiler_opt/tools/extract_ir.py
+++ b/compiler_opt/tools/extract_ir.py
@@ -28,6 +28,11 @@ specifying -mllvm -lto-embed-bitcode=post-merge-pre-opt.
 
 In a local ThinLTO case, the compilation is assumedto have been performed
 specifying -Wl,--save-temps=import -Wl,--thinlto-emit-index-files
+
+To change the logging verbosity, pass an integer representing the desired
+verbosity to the --verbosity flag. Use 0 for all logs and detailed debug
+information, -1 for solely warnings, and -2 to only print a status message at
+the end.
 """
 
 import json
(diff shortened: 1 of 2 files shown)


mtrofin:
https://abseil.io/docs/python/guides/logging? has already a command line for verbosity (should be `verbosity`).

boomanaiden154:
Updated to use the absl logging facilities rather than custom plumbing everywhere.

compiler_opt/tools/extract_ir.py
```diff
@@ -79,6 +79,12 @@
     'The section name passed to llvm-objcopy. For ELF object files, the '
     'default .llvmbc is correct. For Mach-O object files, one should use '
     '__LLVM,__bitcode')
+flags.DEFINE_enum(
```
mtrofin, line 82: doesn't absl already have a 'verbosity' flag?

boomanaiden154: Ah, yep. It doesn't seem to be documented outside of https://github.com/abseil/abseil-py/blob/9e543208a72300a4f8677fe725550fe8dc242bac/absl/logging/__init__.py#L165 (and of course the logging documentation page doesn't seem to exist on Github, so I can't open a PR against it), so I've added some information on usage in the docstring. Thanks for pointing this out!

compiler_opt/tools/extract_ir.py
```diff
@@ -128,6 +133,9 @@ def main(argv):
   extract_ir_lib.write_corpus_manifest(FLAGS.thinlto_build,
                                        relative_output_paths, FLAGS.output_dir)
 
+  # Reset logging verbosity so that we print status messages
```
mtrofin, line 136: i'd drop this - if the user doesn't want any status, let's not :)

compiler_opt/tools/extract_ir_lib.py
```diff
@@ -135,9 +135,13 @@ def _extract_clang_artifacts(self, llvm_objcopy_path: str, cmd_filter: str,
       return None
     os.makedirs(self.dest_dir(), exist_ok=True)
     try:
+      subprocess_output = None if logging.get_verbosity(
```
mtrofin, line 138: why?

boomanaiden154: If there is a failure, `llvm-objcopy` will log it to `STDERR`/`STDOUT` and that will go to the standard output/error of the python process if we don't specify this so the user will end up seeing all of the error messages from `llvm-objcopy` despite specifying `--verbosity=-1` or `--verbosity=-2`.

mtrofin: wouldn't a better handling be capturing `llvm-objcopy`'s output and `logging.<whatever>` it?

boomanaiden154: Yep, we can do that too. That would probably be better because it gives us explicit control over the level that we log it at.

boomanaiden154: Should be updated now. I had to refactor to check the return code of the subprocess and fail on that rather than catch an exception in a try except block, but it preserves the functionality and isn't too big of a change.

compiler_opt/tools/extract_ir_lib.py
```diff
@@ -134,32 +134,40 @@ def _extract_clang_artifacts(self, llvm_objcopy_path: str, cmd_filter: str,
...
-          shutil.copy(index_file, self.thinlto_index_file())
-
-      subprocess.run(
-          self._get_extraction_bc_command(llvm_objcopy_path,
-                                          bitcode_section_name),
-          check=True)
-    except subprocess.CalledProcessError as e:
-      # This may happen if  .o file was build from asm (.S source).
-      logging.warning('%s was not processed: %s', self.input_obj(), e)
+    cmd_objcopy_output = subprocess.run(
```
mtrofin, line 137: why not use `check_output` in a try/catch, to avoid explicitly checking the `returncode` (i.e. seems like something that API is for?)

compiler_opt/tools/extract_ir_lib.py
```diff
@@ -134,32 +134,40 @@ def _extract_clang_artifacts(self, llvm_objcopy_path: str, cmd_filter: str,
...
+        shutil.copy(index_file, self.thinlto_index_file())
+
+    bc_objcopy_output = subprocess.run(
+        self._get_extraction_bc_command(llvm_objcopy_path,
+                                        bitcode_section_name),
+        check=False,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.STDOUT,
+        universal_newlines=True)
+    if bc_objcopy_output.returncode != 0:
```
mtrofin, line 168: same

mtrofin (commented):
just some nits

compiler_opt/tools/extract_ir_lib.py
boomanaiden154: Fixed. That is definitely a cleaner solution.

mtrofin (approved):
please don't forget to change the commit message

PR #283, opened 2023-08-02
Make make_corpus functions consistent with extract_ir functions

boomanaiden154:
Currently the functions in make_corpus_lib assume they are working with a full file path whereas the tooling in extract_ir truncates the .bc/.cmd extensions and only has the object file. This discrepancy means that tooling that works with downstream products of corpora from extract_ir doesn't work without modification with corpora from make_corpus (ignoring the missing .cmd files). This patch makes make_corpus match the behavior of extract_ir.

### DIFF
--- a/compiler_opt/tools/make_corpus_lib.py
+++ b/compiler_opt/tools/make_corpus_lib.py
@@ -21,6 +21,8 @@ import json
 
 from typing import List, Optional
 
+BITCODE_EXTENSION = '.bc'
+
 
 def load_bitcode_from_directory(bitcode_base_dir: str) -> List[str]:
   """Finds bitcode files to extract from a given directory.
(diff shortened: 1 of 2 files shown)


compiler_opt/tools/make_corpus_lib.py
```diff
@@ -33,7 +33,7 @@ def load_bitcode_from_directory(bitcode_base_dir: str) -> List[str]:
   Returns an array of paths representing the relative path to the bitcode
   file from the base direcotry.
   """
-  paths = [str(p) for p in pathlib.Path(bitcode_base_dir).glob('**/*.bc')]
+  paths = [str(p)[:-3] for p in pathlib.Path(bitcode_base_dir).glob('**/*.bc')]
```
mtrofin, line 36: nit: to make it more clear where that "3" came from, you can just have something like:
```
extension = '.bc'
... str(p)[:-len(extension)]
```

and maybe also the glob can be ('**/*' + extension)

mtrofin (approved):

PR #290, opened 2023-08-09
Make combine_corpus only look at folders with a corpus description

boomanaiden154:
Currently the combine corpus tooling just looks through all files and folders in a directory, assuming them all to be folders with a corpus_description.json in them. We don't actually throw an error (only logging one), but it makes sense to be more robust and not log an error when we do something like iterate over a logging file. This patch also removes the tensorflow file code from this library which provides a couple benefits. It means we no longer have to load tensorflow to call the tool which is a significant time-savings and developer tooling like pylint works better on the standard library functions.

### DIFF
--- a/compiler_opt/tools/combine_training_corpus_lib.py
+++ b/compiler_opt/tools/combine_training_corpus_lib.py
@@ -16,10 +16,9 @@
 
 import os
 import json
+import glob
 
 from absl import logging
-
-import tensorflow as tf
 
 _FILE_NAME = 'corpus_description.json'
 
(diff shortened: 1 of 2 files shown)


compiler_opt/tools/combine_training_corpus_lib.py
```diff
@@ -16,29 +16,24 @@
...
-
-    if not tf.io.gfile.exists(path):
-      logging.error('%s does not exist.', path)
-      continue
+  corpus_description_glob = os.path.join(root_dir, '*/' + _FILE_NAME)
+  for corpus_description_path in glob.glob(corpus_description_glob):
+    logging.info('processing %s', corpus_description_path)
 
-    with tf.io.gfile.GFile(path, 'r') as f:
+    with open(corpus_description_path, encoding='utf-8') as f:
```
mtrofin, line 34: please revert this bit.

compiler_opt/tools/combine_training_corpus_lib.py
```diff
@@ -50,5 +45,5 @@ def combine_corpus(root_dir: str) -> None:
 
   output_corpus_description['modules'] = module_names
 
-  with tf.io.gfile.GFile(os.path.join(root_dir, _FILE_NAME), 'w') as f:
+  with open(os.path.join(root_dir, _FILE_NAME), 'w', encoding='utf-8') as f:
```
mtrofin, line 48: please revert

mtrofin (commented):
we do want the tf dependency.

compiler_opt/tools/combine_training_corpus_lib.py
```diff
@@ -16,29 +16,24 @@
 
 import os
 import json
+import glob
 
 from absl import logging
 
-import tensorflow as tf
```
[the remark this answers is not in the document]
boomanaiden154: Makes sense. Didn't realize that's what that was being used for. I've switch back to the tensorflow utilities and am using the tensorflow glob instead to maintain all the file system support of the tensorflow implementation.

mtrofin (approved):

PR #304, opened 2023-10-03
Unify scalar type associations

mtrofin:
We have 3 scalar type representations: tensorflow, ctypes, and string names. Not unlikely we may get a fourth, numpy. This patch factors out all the associations into a central place. Modules can build their own maps as needed from it (i.e. name->ctype or dtype->name, etc)

### DIFF
--- a/compiler_opt/rl/log_reader.py
+++ b/compiler_opt/rl/log_reader.py
@@ -61,29 +61,17 @@ import dataclasses
 import json
 import math
 
+from compiler_opt import type_map
 from typing import Any, BinaryIO, Dict, Generator, List, Optional, Union
 import numpy as np
 import tensorflow as tf
 
-_element_type_name_map = {
-    'float': (ctypes.c_float, tf.float32),
-    'double': (ctypes.c_double, tf.float64),
-    'int8_t': (ctypes.c_int8, tf.int8),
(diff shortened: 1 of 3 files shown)


compiler_opt/rl/log_reader.py
```diff
@@ -61,29 +61,17 @@
...
-}
-
 _element_type_name_to_dtype = {
-    name: dtype for name, (_, dtype) in _element_type_name_map.items()
+    name: dtype for name, _, dtype in type_map.TYPE_ASSOCIATIONS
 }
 
 _dtype_to_ctype = {
-    dtype: ctype for _, (ctype, dtype) in _element_type_name_map.items()
+    dtype: ctype for _, ctype, dtype in type_map.TYPE_ASSOCIATIONS
```
boomanaiden154, line 74: Kind of unrelated to this change, but with this refactoring and the internal `_x_to_y` map being the canonical way to do conversions, is there a reason to keep the `convert_dtype_to_ctype` function around given it's only used internally to this module?

boomanaiden154 (approved):
LGTM. More code reuse is pretty much always a good thing.

compiler_opt/rl/log_reader.py
mtrofin: good point.

PR #336, opened 2024-01-27
Remove infrastructure for publishing ml-compiler-opt as a package

boomanaiden154:
This patch removes infrastructure for publishing ml-compiler-opt as a package. This is no longer needed now as all of the core components that are intended for downstream use are upstream within the LLVM monorepo. Packages are already published under the `mlgo-utils` name.


mtrofin:
What's the tradeoff of leaving it as a package depending on `mlgo-utils` - i.e. it's probably still useful to have it as a package for the training stuff maybe?

boomanaiden154:
> What's the tradeoff of leaving it as a package depending on `mlgo-utils` - i.e. it's probably still useful to have it as a package for the training stuff maybe?

There shouldn't be too much of a compromise just leaving it in. I believe the most useful general use components were the corpus extraction tooling. I think it would be much more difficult to use the training stuff outside of the repository through a package (given things like configuration with gin), but some of it might be useful. I'm not aware of anyone interested in using stuff other than the corpus extraction tooling more broadly and thought removing the infrastructure would make things a little bit cleaner. Not much of a cost to leaving it in, but it should also be easy to restore later.

mtrofin:
gin is a good point, lgtm

mtrofin (approved):

PR #343, opened 2024-05-08
[buildbot] TFLite updates

mtrofin:
- move to TFLite 2.16.2
- bump deps' versions accordingly (based on TF bazel workspace)
- support building with custom sysroot
- support pthreadpool self-build

The latter currently requires a hack whereby it gets to be installed
in the sysroot. We need to patch pthreadpool to support `find_packge`.

### DIFF
--- a/buildbot/build_tflite.sh
+++ b/buildbot/build_tflite.sh
@@ -16,37 +16,49 @@
 
 set -e
 
+if [[ -z "${TFLITE_SYSROOT}" ]]; then
+  EXTRA_CMAKE_FLAGS=" "
+else
+  EXTRA_CMAKE_FLAGS="-DCMAKE_SYSROOT=${TFLITE_SYSROOT} "
+fi
+
+echo extra cmake flags: ${EXTRA_CMAKE_FLAGS}
+
 readonly CPUINFO_REPOSITORY="https://github.com/pytorch/cpuinfo"
-readonly CPUINFO_TAG="5e63739504f0f8e18e941bd63b2d6d42536c7d90"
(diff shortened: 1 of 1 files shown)


buildbot/build_tflite.sh
```diff
@@ -127,10 +145,30 @@ ninja -C gemmlowp/src/gemmlowp-build install
 git clone --filter=tree:0 --no-checkout ${ML_DTYPES_REPOSITORY} ml_dtypes/src/ml_dtypes
 git -C ml_dtypes/src/ml_dtypes checkout ${ML_DTYPES_TAG}
 
+# pthreadpool
+# NOTE: currently a hack, because pthreadpool doesn't support find_package.
+# we install in the default install dir, which really means the script must be run
+# under sudo. Works for buildbots. Not ideal elsewhere.
```
boomanaiden154, line 151: This would need doc updates probably?

I'm wondering if we should just install it to the sysroot if that's specified, but otherwise default to the system dependency.

boomanaiden154 (approved):
Seems reasonable enough to me.

We don't need to patch pthreadpool upstream to support `find_package`, we just need to add a [find module](https://cmake.org/cmake/help/latest/manual/cmake-developer.7.html#find-modules), probably on the tensorflow side. There is support for downloading/specifying the pthreadpool source natively within the tensorflow build, but I remember running into issues the last time I hacked on the build. It should probably be converted to a find module and use an external build just like everything else rather than hackily using CMake to build it as part of the TFLite build if not specifying to use the system dep.

PR #347, opened 2024-07-25
Fix np.Inf for numpy 2

boomanaiden154:
This patch replaces instances of np.Inf with np.inf, which is needed for numpy 2 compatibility.


kazutakahirata (approved):
LGTM

PR #357, opened 2024-09-06
Policy combiner

tvmarino:
Combines two tf-agents policies with the given signature spec in get_input_signature.

### DIFF
--- a/compiler_opt/tools/combine_tfa_policies.py
+++ b/compiler_opt/tools/combine_tfa_policies.py
@@ -1,0 +1,30 @@
+from absl import app
+
+import tensorflow as tf
+
+from compiler_opt.rl import policy_saver
+from compiler_opt.tools import combine_tfa_policies_lib as cfa_lib
+
+
+def main(_):
+  expected_signature = cfa_lib.get_input_signature()
+  action_spec = cfa_lib.get_action_spec()
+  policy1_name = input("First policy name: ")
+  policy1_path = input(policy1_name + " path: ")
+  policy2_name = input("Second policy name: ")
+  policy2_path = input(policy2_name + " path: ")
+  policy1 = tf.saved_model.load(policy1_path, tags=None, options=None)
+  policy2 = tf.saved_model.load(policy2_path, tags=None, options=None)
+  combined_policy = cfa_lib.CombinedTFPolicy(
+     tf_policies={policy1_name:policy1, policy2_name:policy2},
+     time_step_spec=expected_signature,
+     action_spec=action_spec
+  )
+  combined_policy_path = input("Save combined policy path: ")
+  policy_dict = {'combined_policy': combined_policy}
+  saver = policy_saver.PolicySaver(policy_dict=policy_dict)
+  saver.save(combined_policy_path)
+
+if __name__ == "__main__": 
+    app.run(main)
+
--- a/compiler_opt/tools/combine_tfa_policies_lib.py
+++ b/compiler_opt/tools/combine_tfa_policies_lib.py
@@ -1,0 +1,176 @@
+from typing import Dict, List, Optional, Tuple
+
+import gin
+import tensorflow as tf
+import hashlib
+
+import tf_agents
+from tf_agents.trajectories import time_step
+from tf_agents.typing import types
+from tf_agents.trajectories import policy_step
+import tensorflow_probability as tfp
+from tf_agents.specs import tensor_spec
+
+
+class CombinedTFPolicy(tf_agents.policies.TFPolicy):
+
+  def __init__(self, *args,
+               tf_policies: Dict[str, tf_agents.policies.TFPolicy],
+               **kwargs):
+    super(CombinedTFPolicy, self).__init__(*args, **kwargs)
+
+    self.tf_policies = []
+    self.tf_policy_names = []
+    for name, policy in tf_policies.items():
+      self.tf_policies.append(policy)
+      self.tf_policy_names.append(name)
+
+    self.expected_signature = self.time_step_spec
+    self.sorted_keys = sorted(self.expected_signature.observation.keys())
+
+    high_low_tensors = []
+    for name in self.tf_policy_names:
+      m = hashlib.md5()
+      m.update(name.encode('utf-8'))
+      high_low_tensors.append(tf.stack([
+          tf.constant(int.from_bytes(m.digest()[8:], 'little'), dtype=tf.uint64),
+          tf.constant(int.from_bytes(m.digest()[:8], 'little'), dtype=tf.uint64)
+          ])
+      )
+    self.high_low_tensors = tf.stack(high_low_tensors)
+
+    m = hashlib.md5()
+    m.update(self.tf_policy_names[0].encode('utf-8'))
+    self.high = int.from_bytes(m.digest()[8:], 'little')
+    self.low = int.from_bytes(m.digest()[:8], 'little')
+    self.high_low_tensor = tf.constant([self.high, self.low], dtype=tf.uint64)
+
+  def _process_observation(self, observation):
+    for name in self.sorted_keys:
+      if name in ['model_selector']:
+        switch_tensor = observation.pop(name)[0]
+        high_low_tensor = switch_tensor
+    
+        tf.debugging.Assert(
+            tf.equal(
+                tf.reduce_any(
+                    tf.reduce_all(
+                        tf.equal(high_low_tensor, self.high_low_tensors), axis=1
+                        )
+                    ),True
+                ),
+                 [high_low_tensor, self.high_low_tensors])
+        return observation, switch_tensor
+
+  def _create_distribution(self, inlining_prediction):
+    probs = [inlining_prediction, 1.0 - inlining_prediction]
+    logits = [[0.0, tf.math.log(probs[1]/(1.0 - probs[1]))]]
+    return tfp.distributions.Categorical(logits=logits)
+
+  def _action(self, time_step: time_step.TimeStep,
+              policy_state: types.NestedTensorSpec,
+              seed: Optional[types.Seed] = None) -> policy_step.PolicyStep:
+    new_observation = time_step.observation
+    new_observation, switch_tensor = self._process_observation(new_observation)
+    updated_step = tf_agents.trajectories.TimeStep(step_type=time_step.step_type,
+                                      reward=time_step.reward,
+                                      discount=time_step.discount,
+                                      observation=new_observation)
+    def f0():
+      return tf.cast(
+          self.tf_policies[0].action(updated_step).action[0], dtype=tf.int64)
+    def f1():
+      return tf.cast(
+          self.tf_policies[1].action(updated_step).action[0], dtype=tf.int64)
+    action = tf.cond(
+        tf.math.reduce_all(
+            tf.equal(switch_tensor, self.high_low_tensor)),
+        f0,
+        f1
+        )
+    return tf_agents.trajectories.PolicyStep(action=action, state=policy_state)
+
+  def _distribution(
+      self, time_step: time_step.TimeStep,
+      policy_state: types.NestedTensorSpec) -> policy_step.PolicyStep:
+    new_observation = time_step.observation
+    new_observation, switch_tensor = self._process_observation(new_observation)
+    updated_step = tf_agents.trajectories.TimeStep(step_type=time_step.step_type,
+                                      reward=time_step.reward,
+                                      discount=time_step.discount,
+                                      observation=new_observation)
+    def f0():
+      return tf.cast(
+          self.tf_policies[0].distribution(updated_step).action.cdf(0)[0],
+          dtype=tf.float32)
+    def f1():
+      return tf.cast(
+          self.tf_policies[1].distribution(updated_step).action.cdf(0)[0],
+          dtype=tf.float32)
+    distribution = tf.cond(
+        tf.math.reduce_all(
+            tf.equal(switch_tensor, self.high_low_tensor)),
+        f0,
+        f1
+        )
+    return tf_agents.trajectories.PolicyStep(
+        action=self._create_distribution(distribution),
+        state=policy_state)
+
+
+
+@gin.configurable()
+def get_input_signature():
+    """Returns the list of features for LLVM inlining to be used in combining models."""
+    # int64 features
+    inputs = dict(
+        (key,tf.TensorSpec(dtype=tf.int64, shape=(), name=key))
+        for key in [
+            "caller_basic_block_count",
+            "caller_conditionally_executed_blocks",
+            "caller_users",
+            "callee_basic_block_count",
+            "callee_conditionally_executed_blocks",
+            "callee_users",
+            "nr_ctant_params",
+            "node_count",
+            "edge_count",
+            "callsite_height",
+            "cost_estimate",
+            "inlining_default",
+            "sroa_savings",
+            "sroa_losses",
+            "load_elimination",
+            "call_penalty",
+            "call_argument_setup",
+            "load_relative_intrinsic",
+            "lowered_call_arg_setup",
+            "indirect_call_penalty",
+            "jump_table_penalty",
+            "case_cluster_penalty",
+            "switch_penalty",
+            "unsimplified_common_instructions",
+            "num_loops",
+            "dead_blocks",
+            "simplified_instructions",
+            "constant_args",
+            "constant_offset_ptr_args",
+            "callsite_cost",
+            "cold_cc_penalty",
+            "last_call_to_static_bonus",
+            "is_multiple_blocks",
+            "nested_inlines",
+            "nested_inline_cost_estimate",
+            "threshold",
+            "is_callee_avail_external",
+            "is_caller_avail_external",
+        ]
+    )
+    inputs.update({'model_selector': tf.TensorSpec(shape=(2,), dtype=tf.uint64, name='model_selector')})
+    return time_step.time_step_spec(inputs)
+
+@gin.configurable()
+def get_action_spec():
+  return tensor_spec.BoundedTensorSpec(
+    dtype=tf.int64, shape=(), name='inlining_decision', minimum=0, maximum=1
+  )
--- a/compiler_opt/tools/combine_tfa_policies_lib_test.py
+++ b/compiler_opt/tools/combine_tfa_policies_lib_test.py
@@ -1,0 +1,115 @@
+"""Tests for the feature_importance_utils.py module"""
+
+from absl.testing import absltest
+
+import tensorflow as tf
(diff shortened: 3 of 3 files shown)


compiler_opt/tools/combine_tfa_policies_lib.py
```diff
@@ -0,0 +1,193 @@
...
+
+    distribution = tf.cond(
+        tf.math.reduce_all(tf.equal(switch_tensor, self.high_low_tensor)), f0,
+        f1)
+    return tf_agents.trajectories.PolicyStep(
+        action=self._create_distribution(distribution), state=policy_state)
+
+
+@gin.configurable()
+def get_input_signature():
```
mtrofin, line 139: this should come in as a parameter from the user, ie, in this case, something the tool passes in.

also, just use the config registry, like here: https://github.com/google/ml-compiler-opt/blob/dc45cf283a412fa596b9627382a0aaadf059881c/compiler_opt/tools/generate_test_model.py#L46

compiler_opt/tools/combine_tfa_policies_lib.py
```diff
@@ -0,0 +1,193 @@
...
+                ])
+  inputs.update({
+      "model_selector":
+          tf.TensorSpec(shape=(2,), dtype=tf.uint64, name="model_selector")
+  })
+  return ts.time_step_spec(inputs)
+
+
+@gin.configurable()
+def get_action_spec():
```
mtrofin, line 191: same idea as above - this is something the ProblemConfiguration should be able to provide.

compiler_opt/tools/combine_tfa_policies.py
```diff
@@ -0,0 +1,47 @@
...
+"""Runs the policy combiner."""
+from absl import app
+
+import tensorflow as tf
+
+from compiler_opt.rl import policy_saver
+from compiler_opt.tools import combine_tfa_policies_lib as cfa_lib
+
+
+def main(_):
```
mtrofin, line 24: use absl flags to get the 2 policy paths in as command line flags. I'd suggest using one multi-string flag (`policies` or like that).

you also need a flag for the output policy path.

compiler_opt/tools/combine_tfa_policies_lib_test.py
```diff
@@ -0,0 +1,149 @@
...
+from tf_agents.trajectories import time_step as ts
+import tf_agents
+from tf_agents.specs import tensor_spec
+from tf_agents.trajectories import policy_step
+import hashlib
+import numpy as np
+
+
+class AddOnePolicy(tf_agents.policies.TFPolicy):
+  """Test policy which adds one to obs feature."""
```
mtrofin, line 30: nit: s/one/1 (or just say "increments") - easier to read (otherwise may think "one what", rather than realize it's the number `1`)

compiler_opt/tools/combine_tfa_policies_lib_test.py
```diff
@@ -0,0 +1,149 @@
...
+    return ()
+
+  def _action(self, time_step, policy_state, seed):
+    observation = time_step.observation['obs'][0]
+    action = tf.reshape(observation + 1, (1,))
+    return policy_step.PolicyStep(action, policy_state)
+
+
+class SubtractOnePolicy(tf_agents.policies.TFPolicy):
+  """Test policy which subtracts one to obs feature."""
```
mtrofin, line 53: same nit as above.

compiler_opt/tools/combine_tfa_policies_lib_test.py
```diff
@@ -0,0 +1,149 @@
...
+from compiler_opt.tools import combine_tfa_policies_lib
+from tf_agents.trajectories import time_step as ts
+import tf_agents
+from tf_agents.specs import tensor_spec
+from tf_agents.trajectories import policy_step
+import hashlib
+import numpy as np
+
+
+class AddOnePolicy(tf_agents.policies.TFPolicy):
```
mtrofin, line 29: can you add a comment that explains that the _distribution, _variables, and _action members are boylerplate coming from tf_agents? (that's... what it is, right?)

compiler_opt/tools/combine_tfa_policies_lib_test.py
```diff
@@ -0,0 +1,149 @@
...
+    policy2 = SubtractOnePolicy()
+    combined_policy = combine_tfa_policies_lib.CombinedTFPolicy(
+        tf_policies={
+            'add_one': policy1,
+            'subtract_one': policy2
+        },
+        time_step_spec=observation_spec,
+        action_spec=action_spec)
+
+    m = hashlib.md5()
```
mtrofin, line 98: nit: you could refactor the calculation here and reuse it between the 2 tests. if you give the function a name like "client_side_model_selector_calculation" or something like that, a reader will "get it" that this is part of the contract with the user.

compiler_opt/tools/combine_tfa_policies_lib.py
```diff
@@ -0,0 +1,193 @@
...
+      high_low_tensors.append(
+          tf.stack([
+              tf.constant(
+                  int.from_bytes(m.digest()[8:], "little"), dtype=tf.uint64),
+              tf.constant(
+                  int.from_bytes(m.digest()[:8], "little"), dtype=tf.uint64)
+          ]))
+    self.high_low_tensors = tf.stack(high_low_tensors)
+
+    m = hashlib.md5()
```
mtrofin, line 59: you could add a comment here referencing the related LLVM side commit: https://github.com/llvm/llvm-project/pull/96276

compiler_opt/tools/combine_tfa_policies.py
```diff
@@ -0,0 +1,82 @@
...
+from compiler_opt.rl import policy_saver
+from compiler_opt.rl import registry
+from compiler_opt.tools import combine_tfa_policies_lib as cfa_lib
+
+_COMBINE_POLICIES_NAMES = flags.DEFINE_multi_string(
+    'policies_names', [],
+    'List in order of policy names for combined policies.')
+_COMBINE_POLICIES_PATHS = flags.DEFINE_multi_string(
+    'policies_paths', [],
+    'List in order of policy paths for combined policies.')
```
mtrofin, line 32: nit: maybe say here and with the policy names that the orders must match? well.. either that or you could do like a trivial parser, like "name:/policy/path" and just split on `:`. Up to you.

compiler_opt/tools/combine_tfa_policies.py
```diff
@@ -0,0 +1,82 @@
...
+_GIN_BINDINGS = flags.DEFINE_multi_string(
+    'gin_bindings', [],
+    'Gin bindings to override the values set in the config files.')
+
+
+def main(_):
+  flags.mark_flag_as_required('policies_names')
+  flags.mark_flag_as_required('policies_paths')
+  flags.mark_flag_as_required('combined_policy_path')
+  assert len(_COMBINE_POLICIES_NAMES.value) == len(
```
mtrofin, line 46: Do you mean to fail here or just have the assert as a readability tool? We ended up not doing `assert` for errors, and rather either throwing ValueError or - and I think it'd be more appropriate here - check and print the error and exit(1).

May be worth spelling out also in the input policy flag docs that currently we support exactly 2.

compiler_opt/tools/combine_tfa_policies.py
```diff
@@ -0,0 +1,82 @@
...
+      _COMBINE_POLICIES_PATHS.value)
+  gin.add_config_file_search_path(
```
mtrofin, line 48: why?

compiler_opt/tools/combine_tfa_policies_lib.py
```diff
@@ -0,0 +1,138 @@
...
+from tf_agents.typing import types
+from tf_agents.trajectories import policy_step
+import tensorflow_probability as tfp
+
+
+class CombinedTFPolicy(tf_agents.policies.TFPolicy):
+  """Policy which combines two target policies."""
+
+  def __init__(self, *args, tf_policies: Dict[str, tf_agents.policies.TFPolicy],
+               **kwargs):
```
mtrofin, line 32: do we need the `kwargs` right now? let's not have it if not (we can always add later)

compiler_opt/tools/combine_tfa_policies_lib.py
```diff
@@ -0,0 +1,138 @@
...
+    self.high_low_tensor = tf.constant([self.high, self.low], dtype=tf.uint64)
+
+  def _process_observation(
+      self, observation: types.NestedSpecTensorOrArray
+  ) -> Tuple[types.NestedSpecTensorOrArray, types.TensorOrArray]:
+    assert "model_selector" in self.sorted_keys
+    high_low_tensor = self.high_low_tensor
+    for name in self.sorted_keys:
+      if name in ["model_selector"]:
+        switch_tensor = observation.pop(name)[0]
```
mtrofin, line 70: can you add a comment why element 0 is what we need here? (wasn't immediately obvious to me)

compiler_opt/tools/combine_tfa_policies_lib.py
```diff
@@ -0,0 +1,138 @@
...
+            tf.equal(
+                tf.reduce_any(
+                    tf.reduce_all(
+                        tf.equal(high_low_tensor, self.high_low_tensors),
+                        axis=1)), True),
+            [high_low_tensor, self.high_low_tensors])
+
+    return observation, high_low_tensor
+
+  def _create_distribution(self, inlining_prediction):
```
mtrofin, line 83: open q: can we document this isn't introducing a new behavior? or is there some way to delegate the implementation? (OK to document and add a TODO if no easy way to delegate. If you go the TODO path, please open an issue and tag it here, i.e. TODO(1234)..)

compiler_opt/tools/combine_tfa_policies_lib.py
```diff
@@ -0,0 +1,138 @@
...
+              seed: Optional[types.Seed] = None) -> policy_step.PolicyStep:
+    new_observation = time_step.observation
+    new_observation, switch_tensor = self._process_observation(new_observation)
+    updated_step = ts.TimeStep(
+        step_type=time_step.step_type,
+        reward=time_step.reward,
+        discount=time_step.discount,
+        observation=new_observation)
+
+    def f0():
```
mtrofin, line 100: nit: add a comment explaining this is because we currently only support 2 policies, and tag a TODO with an issue about it.

compiler_opt/tools/combine_tfa_policies_lib.py
```diff
@@ -0,0 +1,138 @@
...
+      policy_state: types.NestedTensorSpec) -> policy_step.PolicyStep:
+    new_observation = time_step.observation
+    new_observation, switch_tensor = self._process_observation(new_observation)
+    updated_step = ts.TimeStep(
+        step_type=time_step.step_type,
+        reward=time_step.reward,
+        discount=time_step.discount,
+        observation=new_observation)
+
+    def f0():
```
mtrofin, line 124: copy'n'paste of the above observation (current limit of 2... issue... comment...)

compiler_opt/tools/combine_tfa_policies_lib_test.py
```diff
@@ -0,0 +1,151 @@
...
+            'add_one': policy1,
+            'subtract_one': policy2
+        },
+        time_step_spec=observation_spec,
+        action_spec=action_spec)
+
+    model_selector = client_side_model_selector_calculation('add_one')
+
+    state = ts.TimeStep(
+        discount=tf.constant(np.array([0.]), dtype=tf.float32),
```
mtrofin, line 117: nit: can you use non-zero values for the input to the test, use like `42` :). 

0 could mask what happens if you have (1) zero-initialized values that (2) don't actually get passed to the underlying model.

compiler_opt/tools/combine_tfa_policies.py
```diff
@@ -0,0 +1,82 @@
...
+  gin.parse_config_files_and_bindings(
+      _GIN_FILES.value, bindings=_GIN_BINDINGS.value, skip_unknown=False)
+
+  problem_config = registry.get_configuration()
+  expected_signature, action_spec = problem_config.get_signature_spec()
+  expected_signature.observation.update({
+      'model_selector':
+          tf.TensorSpec(shape=(2,), dtype=tf.uint64, name='model_selector')
+  })
+  assert len(_COMBINE_POLICIES_NAMES.value
```
mtrofin, line 59: Could you open an issue to support more than 2 policies, and add a comment here with a TODO(that issue number) that we should at a point support more than 2. That way (1) we don't forget and (2) a contributor could understand what the goal is and help accordingly (otherwise they may wonder "so why 2...")

mtrofin (commented):
this looks good, just mostly nits

compiler_opt/tools/combine_tfa_policies_lib_test.py
tvmarino: Changed the obs to 42, the other parts of the TimeStep are not used for this test.

compiler_opt/tools/combine_tfa_policies_lib.py
tvmarino: I am just using kwargs and args to pass these to the TFPolicy constructor so if they are needed they would be needed by TFPolicy. I am not completely sure on that last part though.

compiler_opt/tools/combine_tfa_policies_lib.py
tvmarino: This could introduce new behavior, e.g., when combining a deterministic policy such as the saved ppo models. Such policies do not return a proper distribution, however, this is not a good behavior IMO as we use distributions in our trajectory collecting code. In general (my) preferred behavior is to always return a discrete distribution over the action space even if that distribution is supported only on a single action, i.e., deterministic distributions will always look like [1, 0] or [0, 1]. Note that this will only create a new behavior when calling distribution. The behavior of action is remains unchanged.

compiler_opt/tools/combine_tfa_policies.py
tvmarino: I could not get the binary to discover common.gin without adding this line. If there is a different way to do this let me know.

compiler_opt/tools/combine_tfa_policies_lib.py
mtrofin: I understand that, my point is that we don't currently make use of it. So let's remove them, and if we ever need to add them back, that's easy to do.

compiler_opt/tools/combine_tfa_policies_lib.py
mtrofin: > Note that this will only create a new behavior when calling distribution. The behavior of action is remains unchanged.

This means `_create_distribution` isn't used for inference? Can we raise an exception then or something that says "not implemented, and intentionally so"?

compiler_opt/tools/combine_tfa_policies.py
mtrofin: try adding common.gin to the list of gin files you pass in.

compiler_opt/tools/combine_tfa_policies.py
boomanaiden154: It seems like we're still missing a TODO here?

compiler_opt/tools/combine_tfa_policies_lib.py
boomanaiden154: Looks like the TODO should be linked here too?

compiler_opt/tools/combine_tfa_policies_lib_test.py
```diff
@@ -0,0 +1,151 @@
...
+    'obs':
+        tf.TensorSpec(dtype=tf.int32, shape=(), name='obs'),
+    'model_selector':
+        tf.TensorSpec(shape=(2,), dtype=tf.uint64, name='model_selector')
+})
+
+action_spec = tensor_spec.TensorSpec(shape=(1,), dtype=tf.int64)
+
+
+class FeatureImportanceTest(absltest.TestCase):
```
boomanaiden154, line 101: This class name doesn't seem particularly relevant and should probably be switched?

boomanaiden154 (commented):
Just nits from my end.

compiler_opt/tools/combine_tfa_policies.py
tvmarino: I tried passing it as part of gin_files but that didn't work, that's why I added it. I can remove it if you want.

compiler_opt/tools/combine_tfa_policies_lib.py
tvmarino: Raising exceptions in tf did not work in a saved model as you experienced the first time you used the combined policy and didn't pass the right model_selector name. I can remove this function and let _distribution just return something trivial or leave it empty (not sure if this would work), however, if I remember correctly every tf policy needs some type of distribution function.

compiler_opt/tools/combine_tfa_policies_lib.py
tvmarino: It's used to pass time_step_spec. I am keeping it.

compiler_opt/tools/combine_tfa_policies_lib.py
tvmarino: Added the issue and a TODO.

compiler_opt/tools/combine_tfa_policies.py
tvmarino: Added the issue and a TODO.

compiler_opt/tools/combine_tfa_policies_lib.py
mtrofin: Ack on there needing to be a function.

Is it correct that the function isn't actually used during inference, and then it's just that we can't just throw (because of how tf evaluates it)... can we make it return something _invalid_ (and document what's going on, too)? Basically, we want to (1) be clear this is just an artifact of the framework (tf_agents), not actually needed, (2) make it fail-fast if our assumptions in (1) are invalidated in the future, which helps both maintaining the code, and a reader of the code.

compiler_opt/tools/combine_tfa_policies.py
mtrofin: You added it when passing in the gin files as a command line param you mean?

tvmarino: Yes. I passed it as --gin_files='compiler_opt/rl/inlining/gin_configs/common.gin'.

compiler_opt/tools/combine_tfa_policies_lib.py
tvmarino: Set a default deterministic distribution as discussed.

mtrofin (approved):

PR #362, opened 2024-09-12
Add a whole corpus sampler

boomanaiden154:
This patch adds a whole corpus sampler. This is intended to be used for trace data collection where we need to compile an entire corpus subset that is extracted using separate tooling in order to generate a reward.


compiler_opt/rl/corpus.py
```diff
@@ -224,6 +224,40 @@ def __call__(self, k: int, n: int = 10) -> List[ModuleSpec]:
     return list(results)
 
 
+class WholeCorpusSampler(Sampler):
+  """Returns the entire corpus every time a sample is requested."""
```
mtrofin, line 228: is the the entire corpus, or it's that subset of IR modules that cover a function graph?

compiler_opt/rl/corpus.py
```diff
@@ -224,6 +224,40 @@ def __call__(self, k: int, n: int = 10) -> List[ModuleSpec]:
...
+  # as super().__init__ is an abstract method that we need to override to
+  # execute it without throwing an error.
+  def __init__(self, module_specs: Tuple[ModuleSpec]):  # pylint: disable=useless-super-delegation
+    super().__init__(module_specs)
+
+  def reset(self):
+    pass
+
+  def __call__(self, k: int, n: int = 10) -> List[ModuleSpec]:
+    """Returns the entire corpus a list of module specs.
```
mtrofin, line 240: nit: is there a comma missing?

compiler_opt/rl/corpus.py
```diff
@@ -224,6 +224,40 @@ def __call__(self, k: int, n: int = 10) -> List[ModuleSpec]:
...
+
+    Args:
+      k: The number of modules to sample. This must be equal to the number
```
mtrofin, line 243: why do we even need "k" here?

compiler_opt/rl/corpus.py
boomanaiden154: It is the subset of IR modules that cover a trace. That is in effect a whole corpus though given the tooling creates a new `corpus_description.json` for the corpus subset.

compiler_opt/rl/corpus.py
boomanaiden154: Missing a word.

compiler_opt/rl/corpus.py
boomanaiden154: We have to include `k` and `n` or otherwise we get lint errors given the base class has those as parameters.

mtrofin: another way to understand the problem, though, would be to say that given an initial module-level corpus, and a set of subsets of modules, you get a new corpus where a corpus element is such a subset of modules.

then "k" counts how many of those subsets.

boomanaiden154: Sure, but I don't plan on supporting multiple entrypoints to begin with. That would also require more stuff passed around here as we would additionally need to pass the trace associated with that entrypoint associated with this corpus subset.

boomanaiden154:
This is unnecessary when using a custom `BlackboxEvaluator`.

PR #364, opened 2024-09-13
Refactor out policy evaluation to separate class

boomanaiden154:
This patch refactors out policy evaluation in blackbox_learner to a new BlackboxEvaluator class so that we can change some details of how things are collected, particularly with regards to sampling and how samples are held.


compiler_opt/es/blackbox_evaluator.py
```diff
@@ -0,0 +1,109 @@
...
+    raise NotImplementedError()
+
+  @abc.abstractmethod
+  def get_rewards(
+      self, results: List[concurrent.futures.Future]) -> List[Optional[float]]:
+    raise NotImplementedError()
+
+
+@gin.configurable
+class SamplingBlackboxEvaluator(BlackboxEvaluator):
```
mtrofin, line 54: don't you want to also have the non-sampling one implemented, i.e. the old behavior?

boomanaiden154: This is the old behavior. It's basically just copy and pasted from `blackbox_learner.py`.

mtrofin (approved):

PR #369, opened 2024-09-16
Explicitly Type CorpusSample in ES

boomanaiden154:
This patch makes blackbox_learner pass a CorpusSample object to compile jobs rather than a list of ModuleSpecs, making things a bit more readable as now we hold a list of CorpusSamples rather than nested lists of ModuleSpecs.


boomanaiden154:
This also shouldn't be necessary with the new `blackbox_evaluator` abstraction.

PR #370, opened 2024-09-16
Refactor test ESWorker to separate class

boomanaiden154:
This doesn't technically remove a circular dependency, but it does remove an awkward one.


boomanaiden154:
This is actually an issue internally given we cannot have a test depending on a test rule.

mtrofin (approved):

PR #377, opened 2024-10-02
[BC] Add an interactive only mode for env.py

tvmarino:
Added interactive only mode for env.py which compiles only using iclang instead of both clang and iclang.

### DIFF
--- a/compiler_opt/rl/env.py
+++ b/compiler_opt/rl/env.py
@@ -290,6 +290,7 @@ def clang_session(
 def _get_clang_generator(
     clang_path: str,
     task_type: Type[MLGOTask],
+    interactive_only: bool = False,
 ) -> Generator[Optional[Tuple[ClangProcess, InteractiveClang]],
                Optional[corpus.LoadedModuleSpec], None]:
   """Returns a generator for creating InteractiveClang objects.
@@ -299,6 +300,7 @@ def _get_clang_generator(
   Args:
     clang_path: Path to the clang binary to use within InteractiveClang.
     task_type: Type of the MLGO task to use.
+    interactive_only: If set to true only iclang is yielded
 
   Returns:
     The generator for InteractiveClang objects.
@@ -311,9 +313,12 @@ def _get_clang_generator(
     module = yield
     with clang_session(
         clang_path, module, task_type, interactive=True) as iclang:
-      with clang_session(
-          clang_path, module, task_type, interactive=False) as clang:
-        yield iclang, clang
+      if interactive_only:
+        yield iclang, iclang
+      else:
+        with clang_session(
+            clang_path, module, task_type, interactive=False) as clang:
+          yield iclang, clang
 
 
 class MLGOEnvironmentBase:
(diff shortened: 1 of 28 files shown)


compiler_opt/rl/env.py
```diff
@@ -311,9 +313,12 @@ def _get_clang_generator(
     module = yield
     with clang_session(
         clang_path, module, task_type, interactive=True) as iclang:
-      with clang_session(
-          clang_path, module, task_type, interactive=False) as clang:
-        yield iclang, clang
+      if interactive_only:
+        yield iclang, iclang
```
mtrofin, line 317: Could we return iclang here, i.e. not a tuple, and detect that on the receiving end? or (iclang, None)?

tvmarino: The relative reward is computed, that is, the compute_relative_rewards function of env.py always expects two scores. I found this to be the cleanest change. Returning only iclang and None will end up being identical but would require multiple changes and I didn't want to break anything.

mtrofin: OK - can you update the documentation string explaining this, too? (i.e. it's not quite "`interactive_only: If set to true only iclang is yielded`" - looking at the docstring, the return isn't quite well specified actually, what the generator generates is what's interesting - could you basically make the doc "right"?)

compiler_opt/rl/env_test.py
```diff
@@ -215,6 +215,32 @@ def test_env(self, mock_popen):
...
+      self.assertEqual(step.step_type, env.StepType.FIRST)
+
+      for step_itr in range(_NUM_STEPS - 1):
+        del step_itr
+        step = test_env.step(np.array([1], dtype=np.int64))
+        self.assertEqual(step.step_type, env.StepType.MID)
+
+      step = test_env.step(np.array([1], dtype=np.int64))
+      self.assertEqual(step.step_type, env.StepType.LAST)
+      self.assertEqual(step.reward, {'default': 0.})
```
mtrofin, line 242: is the only way we know it's interactive by seeing the reward be 0? Can we test that the 2 iclangs are the same, too?

tvmarino: We can directly check if test_env._iclang and test_env._clang are the same object.

compiler_opt/rl/env.py
```diff
@@ -293,17 +293,19 @@ def _get_clang_generator(
...
   Args:
     clang_path: Path to the clang binary to use within InteractiveClang.
     task_type: Type of the MLGO task to use.
-    interactive_only: If set to true only iclang is yielded
+    interactive_only: If set to true the returned tuple of generators is
+      iclang, iclang instead of iclang, clang
 
   Returns:
-    The generator for InteractiveClang objects.
+    A tuple of generators created with clang_session. First argument of
```
mtrofin, line 305: nit: I think the return is a generator of tuples.

mtrofin (approved):
LGTM, nit in the comment

PR #383, opened 2024-10-14
Initial ExplorationWorker commit

tvmarino:
Initial ExplorationWorker commit. ExplorationWorker implements the
interactive clang compilation together with exploration for a
given module. This commit adds the class with its compile_module
function which will compile the module with a given compiler policy.

### DIFF
--- a/compiler_opt/rl/generate_bc_trajectories.py
+++ b/compiler_opt/rl/generate_bc_trajectories.py
@@ -17,6 +17,7 @@
 from typing import Any, Callable, Dict, List, Optional, Tuple, Type
 
 from absl import logging
+import dataclasses
 import os
 import shutil
 
@@ -41,7 +42,8 @@ def add_int_feature(
   Args:
     sequence_example: sequence example to use instead of the one belonging to
       the instance
-    feature_value: tf.int64 value of feature
+    feature_value: np.int64 value of feature, this is the required type by
+      tf.train.SequenceExample for an int list
     feature_name: name of feature
   """
   f = sequence_example.feature_lists.feature_list[feature_name].feature.add()
@@ -59,7 +61,8 @@ def add_float_feature(
   Args:
     sequence_example: sequence example to use instead of the one belonging to
       the instance
-    feature_value: tf.int64 value of feature
+    feature_value: np.float32 value of feature, this is the required type by
+      tf.train.SequenceExample for an float list
     feature_name: name of feature
   """
   f = sequence_example.feature_lists.feature_list[feature_name].feature.add()
@@ -109,6 +112,13 @@ def add_feature_list(seq_example: tf.train.SequenceExample,
     add_function = add_string_feature
   for feature in feature_list:
     add_function(seq_example, feature, feature_name)
+
+
+@dataclasses.dataclass
+class SequenceExampleFeatureNames:
+  action: str = 'action'
+  reward: str = 'reward'
+  module_name: str = 'module_name'
 
 
 class ExplorationWithPolicy:
@@ -219,10 +229,11 @@ class ExplorationWorker(worker.Worker):
       exploration_policy_distr=None,
       obs_action_specs: Optional[Tuple[time_step.TimeStep,
                                        tensor_spec.BoundedTensorSpec,]] = None,
+      reward_key: str = '',
       **kwargs,
   ):
-    self.loaded_module_spec = loaded_module_spec
-    self.use_greedy = use_greedy
+    self._loaded_module_spec = loaded_module_spec
+    self._use_greedy = use_greedy
     if not obs_action_specs:
       obs_spec = None
       action_spec = None
@@ -230,28 +241,28 @@ class ExplorationWorker(worker.Worker):
       obs_spec = obs_action_specs[0].observation
       action_spec = obs_action_specs[1]
 
-    if 'reward_key' not in kwargs:
-      raise KeyError(
+    if reward_key == '':
+      raise TypeError(
           'reward_key not specified in ExplorationWorker initialization.')
-    self.reward_key = kwargs['reward_key']
+    self._reward_key = reward_key
     kwargs.pop('reward_key', None)
-    self.working_dir = None
+    self._working_dir = None
 
-    self.env = env.MLGOEnvironmentBase(
+    self._env = env.MLGOEnvironmentBase(
         clang_path=clang_path,
         task_type=mlgo_task,
         obs_spec=obs_spec,
         action_spec=action_spec,
     )
-    if self.env.action_spec:
-      if self.env.action_spec.dtype != tf.int64:
+    if self._env.action_spec:
+      if self._env.action_spec.dtype != tf.int64:
         raise TypeError(
             f'Environment action_spec type {0} does not match tf.int64'.format(
-                self.env.action_spec.dtype))
-    self.exploration_frac = exploration_frac
-    self.max_exploration_steps = max_exploration_steps
-    self.exploration_policy_distr = exploration_policy_distr
-    logging.info('Reward key in exploration worker: %s', self.reward_key)
+                self._env.action_spec.dtype))
+    self._exploration_frac = exploration_frac
+    self._max_exploration_steps = max_exploration_steps
+    self._exploration_policy_distr = exploration_policy_distr
+    logging.info('Reward key in exploration worker: %s', self._reward_key)
 
   def compile_module(
       self,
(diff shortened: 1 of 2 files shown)


compiler_opt/rl/generate_bc_trajectories.py
```diff
@@ -14,12 +14,101 @@
...
 from tf_agents.trajectories import policy_step
 from tf_agents.trajectories import time_step
+from tf_agents.specs import tensor_spec
+
+from compiler_opt.distributed import worker
+from compiler_opt.rl import corpus
+from compiler_opt.rl import env
+
+
+def add_int_feature(
```
mtrofin, line 34: should this take a `type` parameter defaulted to np.int64? the type of the value should match the type of the element value in the feature list, right?

same q for the `add_float_feature`

compiler_opt/rl/generate_bc_trajectories.py
```diff
@@ -14,12 +14,101 @@
...
+      the instance
+    feature_value: tf.int64 value of feature
+    feature_name: name of feature
+  """
+  f = sequence_example.feature_lists.feature_list[feature_name].feature.add()
+  lst = f.float_list.value
+  lst.extend([feature_value])
+
+
+def add_string_feature(
```
mtrofin, line 70: why, do we ever think we'd use this?

compiler_opt/rl/generate_bc_trajectories.py
```diff
@@ -102,3 +191,151 @@ def get_advice(self, state: time_step.TimeStep) -> np.ndarray:
...
+      mlgo_task: Type[env.MLGOTask],
+      use_greedy: bool = False,
+      exploration_frac: float = 1.0,
+      max_exploration_steps: int = 10,
+      exploration_policy_distr=None,
+      obs_action_specs: Optional[Tuple[time_step.TimeStep,
+                                       tensor_spec.BoundedTensorSpec,]] = None,
+      **kwargs,
+  ):
+    self.loaded_module_spec = loaded_module_spec
```
mtrofin, line 224: isn't the coding convention that the member fields start with `_`?

compiler_opt/rl/generate_bc_trajectories.py
```diff
@@ -102,3 +191,151 @@ def get_advice(self, state: time_step.TimeStep) -> np.ndarray:
...
+    self.use_greedy = use_greedy
+    if not obs_action_specs:
+      obs_spec = None
+      action_spec = None
+    else:
+      obs_spec = obs_action_specs[0].observation
+      action_spec = obs_action_specs[1]
+
+    if 'reward_key' not in kwargs:
```
mtrofin, line 233: why not make `reward_key` an explicit parameter? if you type it appropriately (`str`, not `Optional[str]`), we can get a type checking error if the user doesn't pass it.

compiler_opt/rl/generate_bc_trajectories.py
```diff
@@ -102,3 +191,151 @@ def get_advice(self, state: time_step.TimeStep) -> np.ndarray:
...
+      shutil.rmtree(working_dir_head)
+    if horizon <= 0:
+      raise ValueError(
+          f'Policy did not take any inlining decision for module {0}.'.format(
+              self.loaded_module_spec.name))
+    if curr_obs_dict.step_type != env.StepType.LAST:
+      raise ValueError(
+          f'Compilation loop exited at step type {0} before last step'.format(
+              curr_obs_dict.step_type))
+    native_size = curr_obs_dict.score_policy[self.reward_key]
```
mtrofin, line 298: `native_size` is tied to inlining, perhaps `score`? same then below

compiler_opt/rl/generate_bc_trajectories.py
```diff
@@ -102,3 +191,151 @@ def get_advice(self, state: time_step.TimeStep) -> np.ndarray:
...
+    native_size_list = np.float32(native_size) * np.float32(np.ones(horizon))
+    add_feature_list(sequence_example, native_size_list, 'reward')
```
mtrofin, line 300: so I understand: the reward key could be `hi_mom`, but we rename it here to `reward`?

compiler_opt/rl/generate_bc_trajectories.py
```diff
@@ -102,3 +191,151 @@ def get_advice(self, state: time_step.TimeStep) -> np.ndarray:
...
+    module_name_list = [self.loaded_module_spec.name for _ in range(horizon)]
+    add_feature_list(sequence_example, module_name_list, 'module_name')
```
mtrofin, line 302: you may want to pull up in a global variable the feature names that are basically forming an API: IIUC `module_name` (1) can't be a feature in the original policy, and (2) is used as part of the implementation here.

You can also validate then that the given trace doesn't have this feature

compiler_opt/rl/generate_bc_trajectories.py
```diff
@@ -102,3 +191,151 @@ def get_advice(self, state: time_step.TimeStep) -> np.ndarray:
...
+    self.exploration_frac = exploration_frac
+    self.max_exploration_steps = max_exploration_steps
+    self.exploration_policy_distr = exploration_policy_distr
+    logging.info('Reward key in exploration worker: %s', self.reward_key)
+
+  def compile_module(
+      self,
+      policy: Callable[[Optional[time_step.TimeStep]], np.ndarray],
+  ) -> tf.train.SequenceExample:
+    """Compiles the module with the given policy and outputs a seq. example.
```
mtrofin, line 260: the documentation should capture that the sequence example also has some special, specifically-named features (`module_name`...) - IIUC

compiler_opt/rl/generate_bc_trajectories.py
```diff
@@ -102,3 +191,151 @@ def get_advice(self, state: time_step.TimeStep) -> np.ndarray:
...
+      reward = np.array(curr_obs_dict.score_policy[self.reward_key])
+    else:
+      reward = np.array(0.)
+    curr_obs_step = step_type_converter[curr_obs_step]
+    timestep = time_step.TimeStep(
+        step_type=tf.convert_to_tensor([curr_obs_step],
+                                       dtype=tf.int32,
+                                       name='step_type'),
+        reward=tf.convert_to_tensor([reward], dtype=tf.float32, name='reward'),
+        discount=tf.convert_to_tensor([0.0], dtype=tf.float32, name='discount'),
```
mtrofin, line 323: `reward` and `discount` here are features required by `TimeStep`, right? one thing that could clarify what feature names mean what is to maybe group them as static members into a couple of classes - just so there's a namespace-ing that helps readability (i.e. here it would be `name=TFAgentsTimeStep.DiscountFeatureName`)

Or, since there don't seem to be many names, just having globals and grouping them with comments also works.

Either way the reader should get sufficient hinting where these names come from or what the intent is with them.

mtrofin (commented):
ptal at your commit message, probably formatting split it.

compiler_opt/rl/generate_bc_trajectories.py
tvmarino: The type value should match the type of the element value in the feature list. All of these are helper functions for add_feature_list. I can definitely make this take a type parameter but I don't see how this would help since at the end of the day I would have to cast whatever type is specified there to an int64 so that we create an int64 feature list. My understanding is that there are only a few types supported for feature lists and the three different add functions are these types.

compiler_opt/rl/generate_bc_trajectories.py
tvmarino: Imagine that during training we want to have specific weights only specific modules, then we can read off the module name from here and compute the relevant weight. It's something that we experimented with, however, it did not significantly improve on our final method. I can imagine this being useful in similar contexts though.

compiler_opt/rl/generate_bc_trajectories.py
tvmarino: The sequence_examples are what is going to be used for training. reward_key is the the key to be used in the environment to determine what we use as a reward, the two are somewhat independent.

compiler_opt/rl/generate_bc_trajectories.py
tvmarino: done

compiler_opt/rl/generate_bc_trajectories.py
tvmarino: done.

compiler_opt/rl/generate_bc_trajectories.py
mtrofin: ah, I see. sgtm

compiler_opt/rl/generate_bc_trajectories.py
mtrofin: if we don't currently (as in, part of this patch or known upcoming ones) use it, let's not add it. It's easy to add later.

compiler_opt/rl/generate_bc_trajectories.py
mtrofin: may be worth commenting that, easier for a reader to get that "it's intentional"

compiler_opt/rl/generate_bc_trajectories.py
mtrofin: just to make things easier for a reader. Rather than realizing that these may be tfagents things (not the most discoverable thing) and then reading on that, 1-2 sentences allows me to "get it" and move forward.

compiler_opt/rl/generate_bc_trajectories.py
tvmarino: I have used it in part of the training for some runs just not the ones that ended up giving the best policy. It would be more effort to remove this and the module_name feature than actually keep it at this point but really up to you.

compiler_opt/rl/generate_bc_trajectories.py
mtrofin: aaah! one thing I was missing: SequenceExamples really knows int64 and float32 IIRC, right? so the constraint here is what SequenceExamples knows about. 

May be worth stressing that somehow. `add_float_feature_to_seq_example` may be a mouthful, maybe just documenting that is sufficient (in the docstring) (i.e. that "int" is really int64 and always will be, because SeqExample. Then the reader gets it that it's not a point in time choice we made; the choice was made for us)

tvmarino: Yeah exactly, I will add something to the documentation to emphasize this.

compiler_opt/rl/generate_bc_trajectories.py
tvmarino: done.

compiler_opt/rl/generate_bc_trajectories.py
tvmarino: changed it to reward

compiler_opt/rl/generate_bc_trajectories.py
tvmarino: See if the docstring for the function makes this more clear now.

compiler_opt/rl/generate_bc_trajectories.py
mtrofin: ok! just wanted to make sure it has an envisioned use (other than test)

compiler_opt/rl/generate_bc_trajectories.py
tvmarino: Check if this looks good, otherwise I haven't understood what you mean here.

tvmarino: Check if the feature names already appear in what is returned by the env is done in _process_obs now.

compiler_opt/rl/generate_bc_trajectories.py
```diff
@@ -111,6 +114,13 @@ def add_feature_list(seq_example: tf.train.SequenceExample,
     add_function(seq_example, feature, feature_name)
 
 
+@dataclasses.dataclass
+class SequenceExampleFeatureNames:
```
mtrofin, line 118: nit: add a comment or doc string saying what this is for ("feature names that we depend on" for example).

also use these in the test

mtrofin (approved):

compiler_opt/rl/generate_bc_trajectories.py
tvmarino: done.

PR #384, opened 2024-10-18
explore_function part of ExplorationWorker

tvmarino:
Explores the module using the given policy and the exploration distr.
This implementation assumes the action set is only {0,1} and will just
switch the action played at explore_step.

### DIFF
--- a/compiler_opt/rl/generate_bc_trajectories.py
+++ b/compiler_opt/rl/generate_bc_trajectories.py
@@ -17,6 +17,7 @@
 from typing import Any, Callable, Dict, List, Optional, Tuple, Type
 
 from absl import logging
+import dataclasses
 import os
 import shutil
 
@@ -41,7 +42,8 @@ def add_int_feature(
   Args:
     sequence_example: sequence example to use instead of the one belonging to
       the instance
-    feature_value: tf.int64 value of feature
+    feature_value: np.int64 value of feature, this is the required type by
+      tf.train.SequenceExample for an int list
     feature_name: name of feature
   """
   f = sequence_example.feature_lists.feature_list[feature_name].feature.add()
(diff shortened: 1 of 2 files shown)


compiler_opt/rl/generate_bc_trajectories.py
```diff
@@ -32,6 +33,30 @@
 from compiler_opt.rl import env
 
 
+@dataclasses.dataclass
+class SequenceExampleFeatureNames:
```
mtrofin, line 37: do you need to rebase?

mtrofin: ok, but to keep the change scoped, can you move it back where it was?

compiler_opt/rl/generate_bc_trajectories.py
```diff
@@ -32,6 +33,30 @@
...
+  """Feature names for features that are always added to seq example."""
+  action: str = 'action'
+  reward: str = 'reward'
+  module_name: str = 'module_name'
+
+
+def get_loss(seq_example: tf.train.SequenceExample,
```
mtrofin, line 44: neat, I wonder if we should move the sequence example stuff in its own utility library? if you think that's a good idea, can you tag an issue on this, we can do it after.

compiler_opt/rl/generate_bc_trajectories.py
```diff
@@ -154,13 +172,15 @@ def __init__(
...
                                                        bool]]] = None,
   ):
-    self.replay_prefix = replay_prefix
-    self.policy = policy
-    self.explore_policy = explore_policy
-    self.curr_step = 0
-    self.explore_step = 0
-    self.gap = np.inf
-    self.explore_on_features = explore_on_features
+    self.explore_step: int = len(replay_prefix) - 1
```
mtrofin, line 175: `self._explore_step`

mtrofin: ah, OK. I see what is also happening. Can you separate the stylistic changes in their own PR, and just submit it once it passes the CI (no review needed, I suggest naming it like "[NFC] add `_` for member fields"). Then merge this.

This helps both with this review and later: the mechanical change doesn't interfere (and one knows not to spend much time looking at it in the future, when we want to understand something)

mtrofin (approved):

PR #387, opened 2024-10-25
[BC] ModuleWorker and ModuleWorkerUtility

tvmarino:
ModuleWorker is responsible for distributing the ExplorationWorker workloads and post-processing the ExplorationWorker.explore_function results. ModuleWorkerUtility has the post-processing utilities used in ModuleWorker.

### DIFF
--- a/compiler_opt/rl/generate_bc_trajectories.py
+++ b/compiler_opt/rl/generate_bc_trajectories.py
@@ -145,7 +145,7 @@ def add_feature_list(seq_example: tf.train.SequenceExample,
 
 def policy_action_wrapper(tf_policy) -> Callable[[Any], np.ndarray]:
   """Return a wrapper for a loaded policy action.
-  
+
     The returned function maps from an (optional) state to an np.array
       that represents the action.
 
--- a/compiler_opt/rl/generate_bc_trajectories.py
+++ b/compiler_opt/rl/generate_bc_trajectories.py
@@ -650,7 +650,7 @@ class ModuleWorker(worker.Worker):
     if exploration_policy_paths:
       if len(exploration_policy_paths) != len(policy_paths):
         raise AssertionError(
-            f'Number of exploration policies: {0}, needs to match number of policies: {1}'
+            f'Number of exploration policies: {0}, doesn't match number of policies: {1}'
             .format(len(exploration_policy_paths), len(policy_paths)))
       for exploration_policy_path in exploration_policy_paths:
         expl_policy = tf.saved_model.load(
--- a/compiler_opt/rl/generate_bc_trajectories.py
+++ b/compiler_opt/rl/generate_bc_trajectories.py
@@ -580,10 +580,8 @@ class ModuleWorkerResultProcessor:
   def __init__(self, base_path: Optional[str] = None):
     self._base_path = base_path
 
-  def partition_for_loss(self,
-                         seq_example: tf.train.SequenceExample,
-                         partitions: List[float],
-                         label_name: str = 'label'):
+  def _partition_for_loss(self, seq_example: tf.train.SequenceExample,
+                          partitions: List[float], label_name: str):
     """Adds a feature to seq_example to partition the examples into buckets.
 
     Given a tuple of partition limits (a_1, a_2, ..., a_n) we create n+1
@@ -602,7 +600,11 @@ class ModuleWorkerResultProcessor:
     add_feature_list(seq_example, label_list, label_name)
 
   def process_succeeded(
-      self, succeeded: List[Tuple[List, List[str], int, float]], spec_name: str
+      self,
+      succeeded: List[Tuple[List, List[str], int, float]],
+      spec_name: str,
+      partitions: List[float],
+      label_name: str = 'label'
   ) -> Tuple[tf.train.SequenceExample, Dict[str, Union[str, float, int]], Dict[
       str, Union[str, float, int]]]:
     seq_example_list = [exploration_res[0] for exploration_res in succeeded]
@@ -632,6 +634,8 @@ class ModuleWorkerResultProcessor:
       temp_working_dir_list = working_dir_list[best_policy_idx][0]
       temp_working_dir = temp_working_dir_list[temp_working_dir_idx]
       self._save_binary(self._base_path, spec_name, temp_working_dir)
+
+    self._partition_for_loss(seq_example, partitions, label_name)
 
     return seq_example, module_dict_max, module_dict_pol
 
@@ -791,9 +795,7 @@ class ModuleWorker(worker.Worker):
                  time_call_compiler)
     (seq_example, module_dict_max,
      module_dict_pol) = self._mw_utility.process_succeeded(
-         succeeded, loaded_module_spec.name)
-
-    self._mw_utility.partition_for_loss(seq_example, self._partitions)
+         succeeded, loaded_module_spec.name, self._partitions)
 
     working_dir_list = [exploration_res[1] for exploration_res in succeeded]
 
--- a/compiler_opt/rl/generate_bc_trajectories.py
+++ b/compiler_opt/rl/generate_bc_trajectories.py
@@ -146,7 +146,7 @@ def add_feature_list(seq_example: tf.train.SequenceExample,
 def policy_action_wrapper(tf_policy) -> Callable[[Any], np.ndarray]:
   """Return a wrapper for a loaded policy action.
 
-    The returned function maps from an (optional) state to an np.array
+  The returned function maps from an (optional) state to an np.array
       that represents the action.
 
   Args:
@@ -168,7 +168,7 @@ def policy_distr_wrapper(
               policy_step.PolicyStep]:
   """Return a wrapper for a loaded tf policy distribution.
 
-    The returned function maps from a state to a distribution over all actions.
+  The returned function maps from a state to a distribution over all actions.
 
   Args:
     tf_policy: A loaded tf policy.
@@ -275,7 +275,7 @@ class ExplorationWithPolicy:
     return policy_action
 
 
-class ExplorationWorker(worker.Worker):
+class ExplorationWorker:
   """Class which implements the exploration for the given module.
 
   Attributes:
@@ -595,15 +595,15 @@ class ModuleWorkerUtility:
       label_name: name of the feature which will contain the bucket index."""
     seq_loss = get_loss(seq_example)
 
-    def find_idx(h, partition):
-      l = 0
-      r = len(partition) - 1
-      while l < r:
-        if h < partition[(l + r) // 2]:
-          r = (l + r) // 2
+    def find_idx(element, partition):
+      left_lim = 0
+      right_lim = len(partition) - 1
+      while left_lim < right_lim:
+        if element < partition[(left_lim + right_lim) // 2]:
+          right_lim = (left_lim + right_lim) // 2
         else:
-          l = (l + r) // 2 + 1
-      idx = r if h < partition[r] else r + 1
+          left_lim = (left_lim + right_lim) // 2 + 1
+      idx = right_lim if element < partition[right_lim] else right_lim + 1
       return idx
 
     label = find_idx(seq_loss, partitions)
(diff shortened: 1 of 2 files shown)


compiler_opt/rl/generate_bc_trajectories.py
```diff
@@ -529,3 +571,248 @@ def _process_obs(self, curr_obs, sequence_example):
...
+    save_dir = os.path.join(base_path, path_head)
+    if not os.path.exists(save_dir):
+      os.makedirs(save_dir, exist_ok=True)
+    shutil.copy(
+        os.path.join(binary_path, 'comp_binary'),
+        os.path.join(save_dir, path_tail))
+
+
+@gin.configurable
+class ModuleWorker(worker.Worker):
```
boomanaiden154, line 685: It's a bit weird that both `ExplorationWorker` and `ModuleWorker` inherit from `worker.Worker`. It seems like only `ModuleWorker` should inherit from `ExplorationWorker` given this one just wraps `ModuleWorker` and `ExplorationWorker` isn't used on its own?

compiler_opt/rl/generate_bc_trajectories.py
```diff
@@ -529,3 +571,248 @@ def _process_obs(self, curr_obs, sequence_example):
...
+        exploration_res = exploration_worker.explore_function(
+            policy_action, explore_policy)
+      except Exception as e:  # pylint: disable=broad-except
+        logging.info('Compilation exception %s at %s', e,
+                     loaded_module_spec.name)
+      if exploration_res is not None:
+        succeeded.append(exploration_res)
+
+    end = timeit.default_timer()
+    time_call_compiler += end - start
```
boomanaiden154, line 798: Is there a reason we need to be recording this here? It seems like something intended for debugging that should probably be pulled out.

boomanaiden154 (commented):
Some initial comments.

compiler_opt/rl/generate_bc_trajectories.py
tvmarino: When collecting a dataset I often found it useful to keep track of how long a module is taking to compile as this gives me a rough idea of how long the whole corpus compilation is going to take. I can of course remove this but will likely end up adding it again for personal use the next time we want to train a new model with BC.

compiler_opt/rl/generate_bc_trajectories.py
tvmarino: That's probably right. I will make the change.

compiler_opt/rl/generate_bc_trajectories.py
boomanaiden154: Ack.

Could we pass this around in a different manner? Maybe just a `logging.info` call after computing the time taken? I can't see how you're planning on using it, but I think it would be a bit nicer if we can avoid passing this around.

compiler_opt/rl/generate_bc_trajectories.py
```diff
@@ -139,6 +143,46 @@ def add_feature_list(seq_example: tf.train.SequenceExample,
...
+  return wrap_function
+
+
+def policy_distr_wrapper(
+    tf_policy: policies.TFPolicy
+) -> Callable[[time_step.TimeStep, Optional[tf_types.NestedTensor]],
+              policy_step.PolicyStep]:
+  """Return a wrapper for a loaded tf policy distribution.
+
+    The returned function maps from a state to a distribution over all actions.
```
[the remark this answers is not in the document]
tvmarino: Done.

compiler_opt/rl/generate_bc_trajectories.py
```diff
@@ -139,6 +143,46 @@ def add_feature_list(seq_example: tf.train.SequenceExample,
     add_function(seq_example, feature, feature_name)
 
 
+def policy_action_wrapper(tf_policy) -> Callable[[Any], np.ndarray]:
+  """Return a wrapper for a loaded policy action.
+
+    The returned function maps from an (optional) state to an np.array
```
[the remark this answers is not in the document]
tvmarino: Done.

compiler_opt/rl/generate_bc_trajectories.py
```diff
@@ -529,3 +571,248 @@ def _process_obs(self, curr_obs, sequence_example):
...
+      idx = r if h < partition[r] else r + 1
+      return idx
+
+    label = find_idx(seq_loss, partitions)
+    horizon = len(seq_example.feature_lists.feature_list['action'].feature)
+    label_list = [label for _ in range(horizon)]
+    add_feature_list(seq_example, label_list, label_name)
+
+  def process_succeeded(
+      self, succeeded, spec_name
```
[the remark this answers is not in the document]
tvmarino: Done.

compiler_opt/rl/generate_bc_trajectories.py
tvmarino: Oh, yeah definitely. I think I was using it exactly for logging in the main function which creates a ModuleWorker and process the results from that but it's probably best to have it here.

compiler_opt/rl/generate_bc_trajectories.py
```diff
@@ -529,3 +571,248 @@ def _process_obs(self, curr_obs, sequence_example):
...
+    bucekts with limits [0,a_1), [a_1, a_2), ..., [a_n-1, a_n), [a_n, +infty).
+    The i-th bucket contains all sequence examples with loss in (a_i-1, a_i].
+
+    Args:
+      seq_example: sequence example from the compiled module
+      partitions: a tuple of limits defining the buckets
+      label_name: name of the feature which will contain the bucket index."""
+    seq_loss = get_loss(seq_example)
+
+    def find_idx(h, partition):
```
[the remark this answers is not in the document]
tvmarino: Check if this makes more sense. This is binary search in a sorted list.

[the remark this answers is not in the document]
boomanaiden154: Sorry for not thinking of it in the first place, but this seems like something built in to Python will be able to do?

https://docs.python.org/3.11/library/bisect.html#examples

compiler_opt/rl/generate_bc_trajectories.py
```diff
@@ -529,3 +571,249 @@ def _process_obs(self, curr_obs, sequence_example):
...
+        'module_name':
+            module_name,
+        'loss':
+            float(get_loss(feature_list)),
+        'horizon':
+            len(feature_list.feature_lists.feature_list['action'].feature),
+    }
+    return per_module_dict
+
+  def _save_binary(self, base_path, save_path, binary_path):
```
boomanaiden154, line 672: These parameters need type annotations.

compiler_opt/rl/generate_bc_trajectories.py
```diff
@@ -139,6 +143,46 @@ def add_feature_list(seq_example: tf.train.SequenceExample,
     add_function(seq_example, feature, feature_name)
 
 
+def policy_action_wrapper(tf_policy) -> Callable[[Any], np.ndarray]:
```
boomanaiden154, line 146: Type annotation for the parameter here?

compiler_opt/rl/generate_bc_trajectories.py
tvmarino: I changed ExplorationWorker to not inherit from Worker. I don't see why ModuleWorker should inherit from ExplorationWorker.

compiler_opt/rl/generate_bc_trajectories.py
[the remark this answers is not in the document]
tvmarino: Thanks for pointing me to this, going to use it.

compiler_opt/rl/generate_bc_trajectories.py
```diff
@@ -529,3 +571,249 @@ def _process_obs(self, curr_obs, sequence_example):
           curr_obs_feature, dtype=obs_dtype, name=curr_obs_feature_name)
       add_feature_list(sequence_example, curr_obs_feature,
                        curr_obs_feature_name)
+
+
+class ModuleWorkerUtility:
```
mtrofin, line 577: Nit: Names like "utility" or "helper" never convey much information. How about instead `ModuleWorkerResultProcessor` or `Filter` or `Saver`? (or maybe there's a better name)?

compiler_opt/rl/generate_bc_trajectories.py
```diff
@@ -529,3 +571,249 @@ def _process_obs(self, curr_obs, sequence_example):
...
+  """Utility class to process ExplorationWorker results for ModuleWorker."""
+
+  def __init__(self, base_path: Optional[str] = None):
+    self._base_path = base_path
+
+  def partition_for_loss(self,
```
mtrofin, line 583: does this need to be public, or is it always what happens after `process_succeeded`?

compiler_opt/rl/generate_bc_trajectories.py
mtrofin: Why does ExplorationWorker not need to be a Worker? Put differently, should it have a different suffix then to avoid confusion?

compiler_opt/rl/generate_bc_trajectories.py
tvmarino: It's getting called in ModuleWorker, not sure how I can make it private.

compiler_opt/rl/generate_bc_trajectories.py
tvmarino: I am not sure if it needs to be a worker. I think originally it was used as a worker but then I needed the utility from ModuleWorker and this is why it was inheriting from Worker. I can change the name.

compiler_opt/rl/generate_bc_trajectories.py
tvmarino: Changed the name.

compiler_opt/rl/generate_bc_trajectories.py
mtrofin: is it basically an implementation detail of `process_succeeded`?

compiler_opt/rl/generate_bc_trajectories.py
mtrofin: so it's not meant to be run in a potentially distributed set up?

compiler_opt/rl/generate_bc_trajectories.py
tvmarino: We could potentially move it to ModuleWorker and make it private there.

compiler_opt/rl/generate_bc_trajectories.py
tvmarino: Okay, that's why I said I am not sure, the way it works right now is ModuleWorker is the one that creates all the distributed tasks and each one is handled by an ExploreModule/ExplorationWorker. Which class needs to inherit from Worker in this case? The way I understood @boomanaiden154's comment was that only ModuleWorker needs to be a Worker, is it the opposite?

boomanaiden154: It was my understanding (originally) that `ExplorationWorker` got distributed and just called into `ModuleWorker`. The class actually getting distributed is what needs to inherit from `worker.Worker`, assuming you're using the distributed training setup.

If we're just creating one `ExplorationWorker` at a time, do we maybe want to rename it to something else like `ExplorationManager` to try and avoid confusion with workers?

compiler_opt/rl/generate_bc_trajectories.py
```diff
@@ -231,13 +276,11 @@ def get_advice(self, state: time_step.TimeStep) -> np.ndarray:
     return policy_action
 
 
-class ExplorationWorker(worker.Worker):
+class ExploreModule:
```
mtrofin, line 279: make it a noun: `ModuleExplorer`

mtrofin (approved):
LGTM, just a naming nit

compiler_opt/rl/generate_bc_trajectories.py
tvmarino: Forgot about that, sorry. Going to fix it now.

PR #389, opened 2024-11-14
[BC] Update `explore_at_state_generator` with new exploration strategy

tvmarino:
Updating the exploration strategy as discussed with @alekh.
The strategy works as follows. At the current exploration state,
the exploration policy will return an array of logits. The logit of
the action currently taken is set to `-np.Inf` and an exploration action
is sampled from the remaining available actions. This update required
updating the tests for `ModuleExplorerWorker`. Additionally, added
`module_explorer_type` to `ModuleWorker` to enable overriding the
exploration strategy by overriding `explore_at_state_generator` of
`ModuleExplorer` class.


alekh (approved):
Looks to be in line with the proposed exploration scheme. Thanks!

Issue #391, opened 2024-11-19
Allow for propagating arguments to ```MLGOTask``` in ```clang_session``` of ```env.py```


tvmarino:
Currently there is no support for propagating arguments to ```MLGOTask``` in ```env.py```. This has the unexpected side effect of not being able to use gin to configure ```MLGOTask``` as soon as the task is spawned by a subprocess created from the main thread.

ricor07:
Hello, I'd like to work on the project by fixing this issue. Can you assign this to me? Thanks

mtrofin:
> Hello, I'd like to work on the project by fixing this issue. Can you assign this to me? Thanks

Done, thanks!

(Note a bunch of us are out until the first week of Jan, so PR reviews may be slow until the 6th or so.)

PR #392, opened 2024-11-20
[BC] Initial fix to #391 and rename generate_bc_trajectories*

tvmarino:
Update names of generate_bc_trajectories* to match the repo naming. Initial fix to Issue #391 by parsing the gin config sting in ModuleExplorer instead of wrapping mlgo_task_type to pass it's arguments directly.


boomanaiden154 (commented):
Is it possible for you to rename the files and fix #391 in separate PRs? It's hard to tease apart the fix to #391 from the refactor with how large the diff is from the latter.

PR #393, opened 2024-11-21
[BC] Multiprocessing fix and change to avoid Issue #391

tvmarino:
The main from generate_bc_trajectories needs to be in a separate file otherwise cloudpickle fails. To avoid Issue #391 we also parse the gin config for ```mlgo_task_type``` in each ```ModuleExplorer``` sub-process.


compiler_opt/rl/generate_bc_trajectories.py
```diff
@@ -742,7 +741,7 @@ def __init__(
       clang_path: str = gin.REQUIRED,
       mlgo_task_type: Type[env.MLGOTask] = gin.REQUIRED,
       policy_paths: List[Optional[str]] = [],
-      exploration_frac: float = gin.REQUIRED,
+      exploration_frac: float = 1.0,
```
mtrofin, line 744: wait why?

compiler_opt/rl/generate_bc_trajectories.py
```diff
@@ -869,7 +868,7 @@ def select_best_exploration(
 def gen_trajectories(
     #  pylint: disable=dangerous-default-value
     data_path: str = gin.REQUIRED,
-    delete_flags: Tuple[str, ...] = gin.REQUIRED,
+    delete_flags: Tuple[str, ...] = (),
```
mtrofin, line 871: why?

mtrofin (approved):
lgtm, curious though why default values had to change?

compiler_opt/rl/generate_bc_trajectories.py
tvmarino: For both default parameter changes -- I realized that these should not be gin.REQUIRED and they have standard default values. For exploration_frac the normal behavior is to set it to 1.0 and not constrain the number of exploration steps, for delete_flags the default behavior should be not to delete any flags.

compiler_opt/rl/generate_bc_trajectories.py
```diff
@@ -341,6 +338,8 @@ def __init__(
           'reward_key not specified in ModuleExplorer initialization.')
     self._reward_key = reward_key
     kwargs.pop('reward_key', None)
+    gin_config_str = kwargs.pop('gin_config_str', None)
+    gin.parse_config(gin_config_str)
```
mtrofin, line 342: add a `gin.clear_config()` here. In test, for example, previous tests that do a `gin.bind_parameter` can mess things up.

separate issue: what we need to do with `gin.bind_parameter` in tests (in a maintainable way)

Issue #394, opened 2024-12-04
Tests setting gin bindings should clean them up


mtrofin:
The title says it all. If they don't get cleaned up, they could pollute other tests' gin config, when running tests from e.g. pytest.

The question is how to do it more maintainably than just "clearing". Like is there some unittest annotation or something like that.

PR #395, opened 2024-12-05
Handle `gin.bind_parameter` in tests

mtrofin:
If a test needs to `gin.bind_parameter` it should also clear that, or make the binding within a `gin.config_scope`, otherwise the bindings may leak to other tests if all the tests are run by e.g. pytest.

This patch uses both approaches: for tests that set up bindings in `setUp` we clear in `tearDown`, else we use the scoped approach.

Issue #394


mtrofin:
Note: tested by re-running PR #393 with the explicit `gin.clear_config()` removed.

tvmarino (approved):
LGTM

PR #399, opened 2024-12-18
Add a test for trainer metrics

boomanaiden154:
This patch adds a test for the metrics in the trainer class to ensure that they are actually set. This patch just directly inspects the state rather than trying to ensure that we log to tensorboard too, but this should be good enough, and is definitely better than what we had before (nothing).


compiler_opt/rl/trainer_test.py
```diff
@@ -131,6 +131,27 @@ def test_training_with_multiple_times(self):
...
+    self.assertEqual(0, test_trainer._data_reward_mean.result().numpy())
+    self.assertEqual(0, test_trainer._num_trajectories.result().numpy())
+
+    dataset_iter = _create_test_data(batch_size=3, sequence_length=3)
+    monitor_dict = {'default': {'test': 1}}
+    test_trainer.train(dataset_iter, monitor_dict, num_iterations=10)
+
+    self.assertEqual(1, test_trainer._data_action_mean.result().numpy())
+    self.assertEqual(2, test_trainer._data_reward_mean.result().numpy())
+    self.assertEqual(0, test_trainer._num_trajectories.result().numpy())
```
mtrofin, line 153: kinda unfortunate this stays 0 - can we make it be something else (different from the first state)

mtrofin (approved):

PR #408, opened 2024-12-19
Bump pylint

boomanaiden154:
This patch bumps pylint to the latest version. This is in preparation for bumping the rest of the deps.


mtrofin (approved):

boomanaiden154:
This will break CI temporarily before #406 lands.

PR #410, opened 2024-12-19
[BC] Inlining for size configs.

tvmarino:
Adding configs for collecting imitation learning trajectories for inlining for size.

### DIFF
--- a/compiler_opt/rl/inlining/gin_configs/imitation_learning.gin
+++ b/compiler_opt/rl/inlining/gin_configs/imitation_learning.gin
@@ -7,11 +7,14 @@ env.InliningForSizeTask.llvm_size_path=''
 
 generate_bc_trajectories_lib.ModuleWorker.clang_path=''
 generate_bc_trajectories_lib.ModuleWorker.mlgo_task_type=@env.InliningForSizeTask
-generate_bc_trajectories_lib.ModuleWorker.policy_paths=[]
+generate_bc_trajectories_lib.ModuleWorker.policy_paths=['']
 generate_bc_trajectories_lib.ModuleWorker.exploration_policy_paths=[]
 generate_bc_trajectories_lib.ModuleWorker.explore_on_features=None
 generate_bc_trajectories_lib.ModuleWorker.base_path=''
-generate_bc_trajectories_lib.ModuleWorker.partitions=[0.,]
+generate_bc_trajectories_lib.ModuleWorker.partitions=[
+    285.0, 376.0, 452.0, 512.0, 571.0, 627.5, 720.0, 809.5, 1304.0, 1832.0,
+    2467.0, 3344.0, 4545.0, 6459.0, 9845.0, 17953.0, 29430.5, 85533.5,
+    124361.0]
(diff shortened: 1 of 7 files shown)


compiler_opt/rl/inlining/imitation_learning_runner.py
```diff
@@ -0,0 +1,49 @@
...
+import gin
+
+from compiler_opt.rl.imitation_learning import generate_bc_trajectories_lib
+from compiler_opt.rl.inlining import imitation_learning_config
+from compiler_opt.tools import generate_test_model  # pylint:disable=unused-import
+
+from tf_agents.system import system_multiprocessing as multiprocessing
+
+flags.FLAGS['gin_files'].allow_override = True
+flags.FLAGS['gin_bindings'].allow_override = True
```
boomanaiden154, line 30: Where are these flags defined? They should be declared within the entrypoint rather than overriden.

boomanaiden154 (commented):
This patch seems to do a lot more than just add some gin configs?

compiler_opt/rl/inlining/imitation_learning_runner.py
tvmarino: I think this is used in the same way as in, e.g., https://github.com/google/ml-compiler-opt/blob/513d50da3662081a0e20ea1222788e8a6718b17d/compiler_opt/tools/generate_default_trace_test.py#L35 
From what I remember I need it because there are gin bindings set by common.gin which I need to overwrite. We had discussed this with Mircea a while back and this was the recommended way to proceed.

tvmarino:
> This patch seems to do a lot more than just add some gin configs?

The patch adds a runner file because not everything can be set in gin files. As discussed, there is a problem with passing functions in gin files which need to be later pickled. The patch adds a runner and a python config file to define the needed functions for the specific inlining problem.

compiler_opt/rl/inlining/gin_configs/imitation_learning.gin
```diff
@@ -0,0 +1,27 @@
...
+
+
+env.InliningForSizeTask.llvm_size_path=''
+
+generate_bc_trajectories_lib.ModuleWorker.clang_path=''
+generate_bc_trajectories_lib.ModuleWorker.mlgo_task_type=@env.InliningForSizeTask
+generate_bc_trajectories_lib.ModuleWorker.policy_paths=[]
+generate_bc_trajectories_lib.ModuleWorker.exploration_policy_paths=[]
+generate_bc_trajectories_lib.ModuleWorker.explore_on_features=None
+generate_bc_trajectories_lib.ModuleWorker.base_path=''
```
mtrofin, line 13: (as discussed offline) let's go with the patch as-is right now, and in a follow-up we remove this from gin and pipe it through from its flag.

mtrofin (approved):

compiler_opt/rl/inlining/imitation_learning_runner.py
tvmarino: This is now fixed, the flags are removed as it turns out they were not needed anymore.

PR #411, opened 2024-12-20
[BC] ```keep_temps``` update

tvmarino:
Rename ```base_path``` to ```persistent_objects_path``` and ```keep_temps``` to ```explicit_temps_dir```.
Makes sure that if ```persistent_objects_path``` is set then ```explicit_temps_dir``` is also set
or neither of them is set. Further, passes ```explicit_temps_dir``` to
```compilation_runner.get_workdir_context``` and ensures that the flag
and argument are not set at the same time.

### DIFF
--- a/compiler_opt/rl/compilation_runner.py
+++ b/compiler_opt/rl/compilation_runner.py
@@ -122,7 +122,7 @@ class ProcessKilledError(Exception):
     Exception.__init__(self)
 
 
-def kill_process_ignore_exceptions(p: 'subprocess.Popen[bytes]'):
+def kill_process_ignore_exceptions(p: 'subprocess.Popen[bytes]'):  # pylint: disable=useless-return
   # kill the process and ignore exceptions. Exceptions would be thrown if the
   # process has already been killed/finished (which is inherently in a race
   # condition with us killing it)
@@ -130,7 +130,7 @@ def kill_process_ignore_exceptions(p: 'subprocess.Popen[bytes]'):
     p.kill()
     p.wait()
   finally:
-    return  # pylint: disable=lost-exception
+    return  # pylint: disable=lost-exception,return-in-finally
 
 
 class WorkerCancellationManager:
--- a/compiler_opt/rl/compilation_runner.py
+++ b/compiler_opt/rl/compilation_runner.py
@@ -39,8 +39,8 @@ COMPILATION_TIMEOUT = flags.DEFINE_integer(
     'Max duration (in seconds) after which we cancel any compilation job.')
 _QUIET = flags.DEFINE_bool(
     'quiet', True, 'Whether or not to compile quietly (hiding info logging)')
-_KEEP_TEMPS = flags.DEFINE_string(
-    'keep_temps', None,
+_EXPLICIT_TEMPS_DIR = flags.DEFINE_string(
+    'explicit_temps_dir', None,
     'Put temporary files into given directory and keep them past exit.')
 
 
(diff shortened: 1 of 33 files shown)

### CODE REVIEW

File: compiler_opt/rl/compilation_runner.py [hunk 1]
```diff
@@ -80,14 +80,23 @@ def __exit__(self, exc, value, tb):
     pass
 
 
-def get_workdir_context():
+def get_workdir_context(keep_temps: Optional[str] = None):
```
mtrofin, line 83: nit: s/keep_temps/explicit_temps_dir. Otherwise it sounds like a boolean (yes, documentation explains, but why call the object bicycle when it's a car lol)

compiler_opt/rl/compilation_runner.py
```diff
@@ -80,14 +80,23 @@ def __exit__(self, exc, value, tb):
...
   """Return a context which manages how the temperory directories are handled.
 
   When the flag keep_temps is specified temporary directories are stored in
   keep_temps.
+
+  Args:
+    keep_temps: Put temporary files into given directory and keep them
+      past exit when compilining
   """
+  if keep_temps and _KEEP_TEMPS.value:
```
mtrofin, line 93: I see... _KEEP_TEMPS... maybe we can refactor this thing's name after. But let's name the API clearly.

Or you can rename both in one swoop after, up to you.

compiler_opt/rl/imitation_learning/generate_bc_trajectories_lib.py
```diff
@@ -46,6 +46,13 @@
 from compiler_opt.distributed import buffered_scheduler
 from compiler_opt.distributed.local import local_worker_manager
 
+_BASE_PATH = flags.DEFINE_string(
```
mtrofin, line 49: maybe _PERSISTENT_OBJECTS_PATH or something - "base" is very generic.

btw, could we "hitch" this path to be a subdirectory of where we save the model?

mtrofin (approved):
lgtm, naming nits, but all can be addressed asynchronously wrt reviews.

compiler_opt/rl/imitation_learning/generate_bc_trajectories_lib.py
```diff
@@ -918,6 +927,10 @@ def gen_trajectories(
     worker_manager_class: A pool of workers hosted on the local machines, each
       in its own process.
   """
+  if (None in (_BASE_PATH.value, FLAGS.keep_temps) and
+      not all(el is None for el in (_BASE_PATH.value, FLAGS.keep_temps))):
+    raise ValueError(('Both flags keep_temps={FLAGS.keep_temps} and'
```
boomanaiden154, line 932: I think the if statement could probably be simpler if written without all the Python syntactic sugar?

```python
if (_BASE_PATH.value is None or FLAGS.keep_temps is None) and (_BASE_PATH.value is not None and FLAGS.keep_temps is not None):
  pass
```

They're both understandable, but I think this is a quite a bit more readable.

compiler_opt/rl/imitation_learning/generate_bc_trajectories_lib.py
```diff
@@ -918,6 +927,10 @@ def gen_trajectories(
...
+                      'base_path={_BASE_PATH.value} should be set or be None'))
```
boomanaiden154, line 933: Is there a reason that both of these have to be set? I think it would make sense to turning on keeping temps if either of the flags is set rather than requiring both.

(Sorry if I missed some conversation about this on the API design side).

tvmarino: I think keep_temps can be set independently of base_path but if base path is set keep_temps should be set, so I will do this check instead.

compiler_opt/rl/imitation_learning/generate_bc_trajectories_lib.py
tvmarino: Not completely sure how we can do this since the code that trains the model is completely separate from this. We could potentially replace this with a sub-directory of where we save the collected trajectories.

PR #419, opened 2025-01-16
Add TraceBlackboxEvaluator

boomanaiden154:
This patch adds TraceBlackboxEvaluator, an evaluator designed for trace
based cost modelling. It implements the BlackboxEvaluator class, special
casing everything that is needed.

### DIFF
--- a/compiler_opt/es/blackbox_evaluator.py
+++ b/compiler_opt/es/blackbox_evaluator.py
@@ -158,4 +158,8 @@ class TraceBlackboxEvaluator(BlackboxEvaluator):
 
     concurrent.futures.wait(
         futures, return_when=concurrent.futures.ALL_COMPLETED)
+    if len(futures) != 1:
+      raise ValueError('Expected to have one result for setting the baseline,'
+                       f' got {len(futures)}')
+
     self._baseline = futures[0].result()
--- a/compiler_opt/es/blackbox_evaluator.py
+++ b/compiler_opt/es/blackbox_evaluator.py
@@ -45,10 +45,18 @@ class BlackboxEvaluator(metaclass=abc.ABCMeta):
   def set_baseline(self) -> None:
     raise NotImplementedError()
 
-  @abc.abstractmethod
   def get_rewards(
       self, results: List[concurrent.futures.Future]) -> List[Optional[float]]:
-    raise NotImplementedError()
+    rewards = [None] * len(results)
+
+    for i in range(len(results)):
+      if not results[i].exception():
+        rewards[i] = results[i].result()
+      else:
+        logging.info('Error retrieving result from future: %s',
+                     str(results[i].exception()))
+
+    return rewards
 
 
 @gin.configurable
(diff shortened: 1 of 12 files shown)


compiler_opt/es/blackbox_evaluator.py
```diff
@@ -42,7 +42,7 @@ def get_results(
     raise NotImplementedError()
 
   @abc.abstractmethod
-  def set_baseline(self) -> None:
+  def set_baseline(self, pool: FixedWorkerPool) -> None:
```
mtrofin, line 45: should the pool be passed in the ctor? it's not like you'd keep changing the pool every time you called set_baseline, or?

compiler_opt/es/blackbox_evaluator.py
```diff
@@ -101,5 +101,62 @@ def get_results(
...
+  def __init__(self, train_corpus: corpus.Corpus,
+               est_type: blackbox_optimizers.EstimatorType, bb_trace_path: str,
+               function_index_path: str):
+    self._train_corpus = train_corpus
+    self._est_type = est_type
+    self._bb_trace_path = bb_trace_path
+    self._function_index_path = function_index_path
+
+    self._has_baseline = False
+    self._baseline = 1
```
mtrofin, line 122: what's the baseline, and why 1?

mtrofin: Also, can't you model _baseline as `Optional[int]` or what, and then _has_baseline can go away because it's just `_baseline is not None`?

compiler_opt/es/blackbox_evaluator.py
```diff
@@ -101,5 +101,62 @@ def get_results(
...
+        action=lambda w, v: w.compile_corpus_and_evaluate(
+            v[0], v[1], v[2], v[3]),
+        jobs=job_args,
+        worker_pool=pool)
+    concurrent.futures.wait(
+        futures, return_when=concurrent.futures.ALL_COMPLETED)
+    return futures
+
+  def set_baseline(self, pool: FixedWorkerPool) -> None:
+    if self._has_baseline:
```
mtrofin, line 143: can we make re-setting the baseline throw, wouldn't it be a bug? the user can just check if it was set

alternatively, could it be set at ctoring time? probably not, but wanted to check

compiler_opt/es/blackbox_evaluator.py
```diff
@@ -101,5 +101,62 @@ def get_results(
...
+      return
+
+    job_args = [(
+        self._train_corpus.module_specs,
```
mtrofin, line 147: could the args be a dict? readability...

compiler_opt/es/blackbox_evaluator.py
boomanaiden154: I had it set up originally where it just gets called every iteration. There's no real reason to do that though, so I've updated it to throw an error if it is called while the baseline is already set.

We need access to the pool to dispatch a job to get the baseline, and we don't have access in the pool in the constructor, so we cannot.

compiler_opt/es/blackbox_evaluator.py
```diff
@@ -101,5 +101,63 @@ def get_results(
...
+
+  def __init__(self, train_corpus: corpus.Corpus,
+               est_type: blackbox_optimizers.EstimatorType, bb_trace_path: str,
+               function_index_path: str):
+    self._train_corpus = train_corpus
+    self._est_type = est_type
+    self._bb_trace_path = bb_trace_path
+    self._function_index_path = function_index_path
+
+    self._has_baseline = False
```
mtrofin, line 121: What does _has_baseline == True mean if self._baseline is None?

compiler_opt/es/blackbox_evaluator.py
```diff
@@ -101,5 +101,63 @@ def get_results(
...
+
+@gin.configurable
+class TraceBlackboxEvaluator(BlackboxEvaluator):
+  """A blackbox evaluator that utilizes trace based cost modelling."""
+
+  def __init__(self, train_corpus: corpus.Corpus,
+               est_type: blackbox_optimizers.EstimatorType, bb_trace_path: str,
+               function_index_path: str):
+    self._train_corpus = train_corpus
+    self._est_type = est_type
```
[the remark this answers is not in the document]
boomanaiden154: Sure. Let me do this in a follow-up though, because I'm just copying the naming from `SamplingBlackboxEvaluator`.

compiler_opt/es/blackbox_evaluator.py
```diff
@@ -101,5 +101,61 @@ def get_results(
...
+    }]
+
+    _, futures = buffered_scheduler.schedule_on_worker_pool(
+        action=lambda w, args: w.compile_corpus_and_evaluate(**args),
+        jobs=job_args,
+        worker_pool=pool)
+
+    concurrent.futures.wait(
+        futures, return_when=concurrent.futures.ALL_COMPLETED)
+    self._baseline = futures[0].result()
```
mtrofin, line 161: futures is guaranteed to have one entry because...?

mtrofin: (one successful entry I should say)

boomanaiden154: I've added an assertion that we have one entry. If the entry is not successful, `.result()` will raise an error, so explicitly capturing that seems a bit redundant.

mtrofin: ok but can't this happen as a legit outcome at runtime?

mtrofin (approved):

compiler_opt/es/blackbox_evaluator.py
boomanaiden154: We should never run into a situation where `len(futures) == 1`. We might run into the future having an exception if we fail to execute the subprocess, but we want to error out on that anyways.

PR #420, opened 2025-01-23
ruff rules for comprehensions and performance

cclauss:
```
% ruff check --select=ASYNC,C4,E,F,PERF,PIE,W \
             --ignore=E722,E731,E741,F401,PERF203,PIE790 \
             --line-length=130 \
             --statistics

11	C402   	[*] unnecessary-generator-dict
11	W293   	[*] blank-line-with-whitespace
 8	PERF401	[ ] manual-list-comprehension
 5	C401   	[*] unnecessary-generator-set
 2	C405   	[*] unnecessary-literal-set
 2	PIE808 	[*] unnecessary-range-start
 1	C413   	[*] unnecessary-call-around-sorted
 1	PERF102	[*] incorrect-dict-iterator
[*] fixable with `ruff check --fix`
```
% [`ruff rule PERF401`](https://docs.astral.sh/ruff/rules/manual-list-comprehension/)
# manual-list-comprehension (PERF401)

Derived from the **Perflint** linter.

Fix is sometimes available.

## What it does
Checks for `for` loops that can be replaced by a list comprehension.

## Why is this bad?
When creating a transformed list from an existing list using a for-loop,
prefer a list comprehension. List comprehensions are more readable and
more performant.

Using the below as an example, the list comprehension is ~10% faster on
Python 3.11, and ~25% faster on Python 3.10.

Note that, as with all `perflint` rules, this is only intended as a
micro-optimization, and will have a negligible impact on performance in
most cases.

## Example
```python
original = list(range(10000))
filtered = []
for i in original:
    if i % 2:
        filtered.append(i)
```

Use instead:
```python
original = list(range(10000))
filtered = [x for x in original if x % 2]
```

If you're appending to an existing list, use the `extend` method instead:
```python
original = list(range(10000))
filtered.extend(x for x in original if x % 2)
```

Take care that if the original for-loop uses an assignment expression
as a conditional, such as `if match:=re.match("\d+","123")`, then
the corresponding comprehension must wrap the assignment
expression in parentheses to avoid a syntax error.

### DIFF
--- a/compiler_opt/benchmark/benchmark_chromium.py
+++ b/compiler_opt/benchmark/benchmark_chromium.py
@@ -206,9 +206,7 @@ def main(_):
     with open(test_description, encoding='UTF-8') as test_description_file:
       print(test_description)
       test_descriptions.append(json.load(test_description_file))
-  test_executables = []
-  for test_description in test_descriptions:
-    test_executables.append(test_description['executable'])
+  test_executables = [test_description['executable'] for test_description in test_descriptions]
 
   if FLAGS.compile_llvm:
     benchmarking_utils.build_llvm(FLAGS.model_path, FLAGS.llvm_use_incremental,
--- a/compiler_opt/benchmark/benchmark_chromium.py
+++ b/compiler_opt/benchmark/benchmark_chromium.py
@@ -206,7 +206,9 @@ def main(_):
     with open(test_description, encoding='UTF-8') as test_description_file:
       print(test_description)
       test_descriptions.append(json.load(test_description_file))
-  test_executables = [test_description['executable'] for test_description in test_descriptions]
+  test_executables = [
+      test_description['executable'] for test_description in test_descriptions
+  ]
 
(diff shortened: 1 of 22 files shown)


mtrofin (approved):

mtrofin:
This is ready to merge

mtrofin:
@cclauss could you "squash & merge" (i.e. checking if that works for you or if you're waiting for us :) I would prefer you merging in case you wanted to update the commit message, etc)

boomanaiden154:
> could you "squash & merge" (i.e. checking if that works for you or if you're waiting for us

People without permissions need someone to click merge on their behalf.

I can do it here once CI passes.

mtrofin:
> > could you "squash & merge" (i.e. checking if that works for you or if you're waiting for us
> 
> People without permissions need someone to click merge on their behalf.

OK, fixed.

> 
> I can do it here once CI passes.

PR #421, opened 2025-01-23
Add ruff to the testing

cclauss:
https://docs.astral.sh/ruff
https://docs.astral.sh/ruff/linter

As suggested at
* https://github.com/google/ml-compiler-opt/pull/420#issuecomment-2610181130

PR #423, opened 2025-01-28
PEP 621: Storing project metadata in pyproject.toml

cclauss:
https://peps.python.org/pep-0621 was approved in 2020.

Migrate project metadata from `.style.yapf` and `pytest.ini` into `pyproject.toml`.

Use [ini2toml](https://pypi.org/project/ini2toml) to convert the pytest file, run [pyproject-fmt](https://pypi.org/project/pyproject-fmt), and then [validate-pyproject](https://github.com/abravalheri/validate-pyproject) to format and validate the results.

Related to:
* #420 
* #421


boomanaiden154 (approved):
Didn't realize this was standard at this point. Seems reasonable enough to me.

PR #425, opened 2025-01-29
misc: ruff rules E741 and YTT204

cclauss:
% [`ruff rule YTT204`](https://docs.astral.sh/ruff/rules/sys-version-info-minor-cmp-int) `sys-version-info-minor-cmp-int`

% [`ruff rule E741`](https://docs.astral.sh/ruff/rules/ambiguous-variable-name)
# ambiguous-variable-name (E741)

Derived from the **pycodestyle** linter.

## What it does
Checks for the use of the characters 'l', 'O', or 'I' as variable names.

Note: This rule is automatically disabled for all stub files
(files with `.pyi` extensions). The rule has little relevance for authors
of stubs: a well-written stub should aim to faithfully represent the
interface of the equivalent .py file as it exists at runtime, including any
ambiguously named variables in the runtime module.

## Why is this bad?
In some fonts, these characters are indistinguishable from the
numerals one and zero. When tempted to use 'l', use 'L' instead.

## Example
```python
l = 0
O = 123
I = 42
```

Use instead:
```python
L = 0
o = 123
i = 42
```

### DIFF
--- a/compiler_opt/distributed/worker.py
+++ b/compiler_opt/distributed/worker.py
@@ -105,7 +105,7 @@ def get_full_worker_args(worker_class: 'type[Worker]', **current_kwargs):
     # it's not a requirement that it were. Tests, for instance, don't use gin.
     pass
   # Issue #38
-  if sys.version_info.minor >= 9:
+  if sys.version_info >= (3, 9):
     return current_kwargs | gin_config
   else:
     return {**current_kwargs, **gin_config}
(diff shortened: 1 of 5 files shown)


cclauss:
Someone who knows Numpy should look at the issues at: `ruff check select=NPY`.

Also, ``ruff check select=E722`
## Why is this bad?
A bare `except` catches `BaseException` which includes
`KeyboardInterrupt`, `SystemExit`, `Exception`, and others. Catching
`BaseException` can make it hard to interrupt the program (e.g., with
Ctrl-C) and can disguise other problems.

mtrofin (approved):

boomanaiden154 (approved):
LGTM.

From https://docs.astral.sh/ruff/rules/sys-version-info-minor-cmp-int/ outlining the motivation for YTT204 since I didn't see a large benefit at first:
> Comparisons based on the current minor version number alone can cause subtle bugs and would likely lead to unintended effects if the Python major version number were ever incremented (e.g., to Python 4).

Malekkimo (commented):
****

PR #428, opened 2025-02-08
[BC] Compiler timeout

tvmarino:
Updates to ```generate_bc_trajectories_lib```, ```env``` and ```log_reader``` to include
timing out the clang processes. Updates to ```weighted_bc_trainer_lib``` to
determine the input shape for the policy network.

### DIFF
--- a/compiler_opt/rl/env.py
+++ b/compiler_opt/rl/env.py
@@ -25,6 +25,9 @@ import contextlib
 import io
 import os
 from typing import Callable, Generator, List, Optional, Tuple, Type
+import threading
+import logging
+import shutil
 
 import numpy as np
 
@@ -227,6 +230,7 @@ def clang_session(
     *,
     explicit_temps_dir: Optional[str] = None,
     interactive: bool,
+    timeout: float = 100.0,
 ):
   """Context manager for clang session.
 
@@ -240,6 +244,7 @@ def clang_session(
     explicit_temps_dir: Put temporary files into given directory and keep them
       past exit when compilining
     interactive: Whether to use an interactive or default clang instance
+    timeout: time in sec after which the clang processes time out
 
   Yields:
     Either the constructed InteractiveClang or DefaultClang object.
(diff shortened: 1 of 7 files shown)


compiler_opt/rl/log_reader.py
mtrofin, line 1: can the change here go in its own PR (with test)?

compiler_opt/rl/inlining/imitation_learning_config.py
```diff
@@ -80,6 +80,7 @@ def get_input_signature():
 @gin.register
 def get_task_type() -> Type[env.InliningForSizeTask]:
   """Returns the task type for the trajectory collection."""
+  #TODO(@tvmarino): make the reward_key match that in ModuleWorker
```
mtrofin, line 83: you can just make a PR with just this and submit it no review required :)

compiler_opt/rl/inlining/gin_configs/imitation_learning.gin
mtrofin, line 1: are the gin changes necessary (in this or any PR)?

compiler_opt/rl/env.py
```diff
@@ -260,14 +265,32 @@ def clang_session(
     def _get_scores() -> dict[str, float]:
       return task.get_module_scores(task_working_dir)
 
+    def delete_working_dir(proc, writer_name, reader_name):
```
mtrofin, line 268: I think this should go in its own PR, maybe it can be the same as the PR for the change to the log_reader (they are quite related, so most probably yes).

You can test by having the fake "clang" the test uses controllably block at each problematic step - you can imagine a few very similar tests, just that each configures the "clang" to fail accordingly.

compiler_opt/rl/log_reader.py
tvmarino: I am not sure what the test should be as I don't want to get into details of log_reader.py. If it's necessary I'll do it but prefer to keep it in this pr.

PR #433, opened 2025-02-09
Move some gradient ascent optimizer flags into the gin config

boomanaiden154:
This patch moves command line flags in es_trainer_lib to set the gradient
ascent optimizer and gradient ascent optimizer flags to the gin config along
with all the other model/optimizer hyperparameters. These don't make a lot of
sense to have as command line parameters given their nature and everything
else similar being defined in gin configs.


compiler_opt/es/es_trainer_lib.py
```diff
@@ -60,11 +51,21 @@
                                      "List of paths to training corpora")
 
 
+@gin.constants_from_enum(module="es_trainer_lib")
+class GradientAscentOptimizerType(enum.Enum):
+  MOMENTUM = 1
```
mtrofin, line 56: Add a `INVALID = 0` maybe

also I think you can use `enum.auto()`

mtrofin (approved):
lgtm, a comment.

PR #435, opened 2025-02-09
Add initial gin configs for regalloc_trace in ES

boomanaiden154:
This patch adds a set of gin configs that can be used for regalloc_trace ES
training. This are intended mostly to be starting points so that we have
a set of options that work upstream as an example and they can be modified
within experiments as needed. They need to be modified before use, such as
by setting binary and basic block trace paths.


mtrofin (approved):
rubber stamp LGTM, I imagine the values are sane.

PR #439, opened 2025-02-11
Test on Python 3.12

cclauss:
What changes will be required to run on Python 3.12?

Blocked by:
* google-deepmind/reverb#145
* tensorflow/agents#952

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -28,11 +28,11 @@ jobs:
         run: |
           if [ -z $ACT ]
           then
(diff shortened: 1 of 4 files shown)


boomanaiden154 (commented):
The lockfile needs to be regenerated when touching `Pipfile`.

This is also blocked on tf-agents.

cclauss:
https://github.com/google/ml-compiler-opt/pull/421#issuecomment-2620066224

Let's close for now and we can revisit this when the two dependencies are Py3.12-ready.

PR #440, opened 2025-02-12
env (interactive mode): timeout for blocking operations

mtrofin:
The `open`ing of the communication named pipes and the
`read` operations are blocking. This introduces a timeout.

High level:
- we associate a thread with `open`, which waits for a signal
  (Event), with a timeout. If it times out, the thread opens
  the associated pipe, which unblocks the main `open`

- writes aren't expected to block
- reads are. To address, we poll (`select.select`) before reading,
  with a timeout.

The design treats the timeout as resetting for each blocking op.
This is just to simplify the design - rather than checking passed
time.


compiler_opt/rl/env.py
```diff
@@ -219,6 +222,135 @@ def _reward_fn(a: float, b: float) -> float:
...
+    with open(filename, 'rb'):
+      pass
+
+  waiter = threading.Thread(target=_timeout_thread)
+  waiter.start()
+  try:
+    with io.BufferedWriter(io.FileIO(filename, 'wb')) as writer_pipe:
+      if not timed_out.is_set():
+        opened.set()
+        yield writer_pipe
```
boomanaiden154, line 253: Why yield here over return? It doesn't seem like we want a generator here.

compiler_opt/rl/env.py
```diff
@@ -219,6 +222,135 @@ def _reward_fn(a: float, b: float) -> float:
...
+          result = wrapped_method(*args, **kwargs)
+          logging.debug('ReaderWithTimeout %s completed', name)
+          return result
+        logging.info('ReaderWithTimeout timed out waiting to %s', name)
+        raise TimeoutError('timed out reading')
+
+      return _replacement
+
+    # pylint: disable=protected-access
+    obj._orig_read = obj.read
```
boomanaiden154, line 294: Why exactly do these need to be put into protected members before being passed to `_replacement(*)`? It doesn't seem like they have any other users.

compiler_opt/rl/env.py
```diff
@@ -219,6 +222,135 @@ def _reward_fn(a: float, b: float) -> float:
...
+    obj.readline = _get_polling_wrapper(obj._orig_readline)
+    obj.readinto = _get_polling_wrapper(obj._orig_readinto)
+    obj.readall = _get_polling_wrapper(obj._orig_readall)
+    # pylint: enable=protected-access
+    return obj
+
+  opened = threading.Event()
+  timed_out = threading.Event()
+
+  # same idea as in the writer case - unblock the `open`
```
boomanaiden154, line 309: Looks like this isn't worth factoring out due to all the differences in logging messages?

boomanaiden154 (commented):
Some initial comments. i'd like to go through the tests a bit more thoroughly when I have chance.

compiler_opt/rl/env.py
boomanaiden154: Ah, looks like it's something that `contextlib.contextmanager` wants.

compiler_opt/rl/env.py
mtrofin: true.

boomanaiden154 (approved):

PR #444, opened 2025-02-15
[BC] Trainer fixes

tvmarino:
Changes to weighted_bc_trainer_lib and weighted_bc_trainer to force only ```_train_step``` to execute in Graph mode. Fix a possible division by 0 in weights creation. only _train_step to execute in Graph mode. Fix a possible division by 0 in weights creation.

### DIFF
--- a/compiler_opt/rl/imitation_learning/generate_bc_trajectories_lib.py
+++ b/compiler_opt/rl/imitation_learning/generate_bc_trajectories_lib.py
@@ -826,7 +826,7 @@ class ModuleWorker(worker.Worker):
       loaded_module_spec: corpus.LoadedModuleSpec,
   ) -> tuple[tuple[int, ProfilingDictValueType, ProfilingDictValueType],
              tf.train.SequenceExample]:
-
+    logging.set_verbosity('info')
     num_calls = len(self._tf_policy_action)
     time_call_compiler = 0
     logging.info('Processing module: %s', loaded_module_spec.name)
--- a/compiler_opt/rl/imitation_learning/weighted_bc_trainer.py
+++ b/compiler_opt/rl/imitation_learning/weighted_bc_trainer.py
@@ -20,6 +20,7 @@ from absl import logging
 import gin
 import json
 from compiler_opt.rl import policy_saver
+import tensorflow as tf
 
 from compiler_opt.rl.inlining import imitation_learning_config as config
 
@@ -77,6 +78,8 @@ def main(_):
   gin.parse_config_files_and_bindings(
       _GIN_FILES.value, _GIN_BINDINGS.value, skip_unknown=False)
   logging.info(gin.config_str())
+
+  tf.compat.v1.enable_eager_execution()
 
   train()
 
--- a/compiler_opt/rl/imitation_learning/weighted_bc_trainer_lib.py
+++ b/compiler_opt/rl/imitation_learning/weighted_bc_trainer_lib.py
@@ -177,8 +177,8 @@ class TrainingWeights:
         bucket_loss += np.maximum(prof[SequenceExampleFeatureNames.regret], 0)
       losses_per_bucket.append(bucket_loss)
     logging.info('Losses per bucket: %s', losses_per_bucket)
-    losses_per_bucket_normalized = losses_per_bucket / np.max(
-        np.abs(losses_per_bucket))
+    losses_per_bucket_normalized = losses_per_bucket / (
+        np.max(np.abs(losses_per_bucket)) + 1e-6)
     probs_t = self._get_exp_gradient_step(losses_per_bucket_normalized, 1.0)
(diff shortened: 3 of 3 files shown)


compiler_opt/rl/imitation_learning/generate_bc_trajectories_lib.py
```diff
@@ -826,7 +826,7 @@ def select_best_exploration(
       loaded_module_spec: corpus.LoadedModuleSpec,
   ) -> tuple[tuple[int, ProfilingDictValueType, ProfilingDictValueType],
              tf.train.SequenceExample]:
-
+    logging.set_verbosity('info')
```
mtrofin, line 829: probably not needed?

mtrofin (approved):

compiler_opt/rl/imitation_learning/weighted_bc_trainer.py
```diff
@@ -13,20 +13,20 @@
...
+import json
 
 import gin
-import json
-from compiler_opt.rl import policy_saver
+import tensorflow as tf
+from absl import app, flags, logging
 
+from compiler_opt.rl import policy_saver
+from compiler_opt.rl.imitation_learning.weighted_bc_trainer_lib import (
```
boomanaiden154, line 23: Nit: all these import changes look superfluous?

compiler_opt/rl/imitation_learning/weighted_bc_trainer_lib.py
```diff
@@ -322,13 +323,12 @@ def load_dataset(self, filepaths: list[str]) -> tf.data.TFRecordDataset:
...
         self._batch_size, drop_remainder=True)  # 4194304
-    dataset = dataset.apply(tf.data.experimental.ignore_errors())
 
     return dataset
 
   def _create_weights(self, labels, weights_arr):
-    p_norm = min(weights_arr)  # check that this should be min
-    weights_arr = tf.map_fn(lambda x: p_norm / x, tf.constant(weights_arr))
+    p_norm = tf.reduce_min(weights_arr)  # check that this should be min
+    weights_arr = tf.map_fn(lambda x: p_norm / x, weights_arr)
```
boomanaiden154, line 331: There's no element-wise division by scalar operation in TF?

Looks like `tf.divide` would do it and is probably much faster than a Python lambda.

boomanaiden154 (commented):
Seems like a portion of the PR description got duplicated?

Also, why only force `_train_step` to execute in graph mode? I would think that we would want everything possible to be in the graph for performance reasons (and to enable things like offloading).

compiler_opt/rl/imitation_learning/generate_bc_trajectories_lib.py
tvmarino: Without this all off the logging inside select_best_exploration is not visible to stdout and I find it useful for e.g. logging the size of the compiled modules during exploration, seeing which module is being currently processed, etc.

compiler_opt/rl/imitation_learning/weighted_bc_trainer.py
tvmarino: This happened after running ruff check --fix I think, otherwise it was complaining about grouping imports.

compiler_opt/rl/imitation_learning/weighted_bc_trainer_lib.py
tvmarino: Will do that, thanks!

tvmarino:
> Seems like a portion of the PR description got duplicated?
> 
> Also, why only force `_train_step` to execute in graph mode? I would think that we would want everything possible to be in the graph for performance reasons (and to enable things like offloading).

The main problem with running everything in graph mode is logging metrics from the train function. I don't seem to be able to do this right now without eager execution for this part of the code. Happy to discuss this more if you have ideas how to refactor the code to still be able to log metrics.

compiler_opt/rl/imitation_learning/weighted_bc_trainer.py
boomanaiden154: Interesting. It definitely shouldn't be doing that. The Google style guide does not allow multiple imports like this (with a few exceptions).

I'll try and take a look at some point. Fine with me to land this as is.

boomanaiden154:
> The main problem with running everything in graph mode is logging metrics from the train function. I don't seem to be able to do this right now without eager execution for this part of the code. Happy to discuss this more if you have ideas how to refactor the code to still be able to log metrics.

Yeah, you won't get any logging if you're entire training loop has been traced (side effects are ignored during tracing). Only enabling graph mode within a train step seems reasonable enough to me. A bit weird that you have to explicitly enable eager execution (it should be enabled by default) to get things working, but should be fine.

Issue #280, opened 2023-07-19
Separate specifics from general-purpose functions in policy_utils


salaast:
The config tests are very problem-specific and should be separated from the general purpose vectorization tests. create_actor_policy should depend less on problem_config registry and instead should just be given the necessary parameters--the signature spec and processing layer creator. Thus, the tests would not depend on specifics.

PR #446, opened 2025-02-20
Refactor policy creation in policy_utils_test

boomanaiden154:
This patch refactors out common policy creation functionality for three
tests into a helper function to get rid of some trivial code
duplication.

Seems like this is a bit of a short term fix for #280.


mtrofin (approved):
lgtm - indeed, can't see a difference between the removed blocks and _save_inlining_policy

PR #447, opened 2025-02-20
Add convert_to_tflite to policy_utils

boomanaiden154:
This patch takes the function that converts policies serialized to numpy
arrays serialized to bytes into TFLite models to policy_utils. This is
necessary so that we can perform this on the worker rather than trying
to perform it serially on the main invocation.


compiler_opt/es/policy_utils_test.py
```diff
@@ -220,6 +220,20 @@ def test_tfpolicy_and_loaded_policy_produce_same_variable_order(self):
...
+
+    output_bytes = self.params.tobytes()
+
+    scratch_dir = self.create_tempdir()
+    tflite_dir = policy_utils.convert_to_tflite(output_bytes, scratch_dir,
+                                                saved_model_path)
+
+    self.assertTrue(os.path.exists(os.path.join(tflite_dir, 'model.tflite')))
+    self.assertTrue(
+        os.path.exists(os.path.join(tflite_dir, 'output_spec.json')))
```
mtrofin, line 235: can you test the resulting policy's weights match what you started with?

boomanaiden154: Not easily. We only have support for loading the parameters from a saved model, not from a TFLite conversion of that saved model.

mtrofin: if you split `convert_to_tflite` at line 150, would it work to test the bit above line 150? The stuff below is fairly standard.

boomanaiden154: I'm not sure it makes too much sense to split. I've added an assertion that the saved model produced during the process has the correct parameters which should have the equivalent effect.

PR #448, opened 2025-02-20
Move TFLite Policy Generation to ES Workers

boomanaiden154:
This patch moves TFLite Policy Generation for ES to the workers. This
step is reasonably expensive and doing it serially on the main node
takes quite some time with a reasonable number of perturbations. This
patch moves this step to the workers to parallelize it.


compiler_opt/es/blackbox_learner.py
```diff
@@ -275,12 +245,16 @@ def run_step(self, pool: FixedWorkerPool) -> None:
...
       ]
 
-    perturbations_as_policies = [
-        self._get_policy_from_perturbation(perturbation)
+    # TODO(boomanaiden154): This should be adding the perturbation to
+    # the existing model weights. That currently results in the model
+    # weights all being NaN, presumably due to rewards not being scaled for
+    # the regalloc_trace problem.
+    perturbations_as_bytes = [
+        perturbation.astype(np.float32).tobytes()
```
mtrofin, line 253: are they always `np.float32`?

compiler_opt/es/regalloc_trace/regalloc_trace_worker.py
```diff
@@ -64,6 +74,12 @@ def __init__(self, clang_path: str, basic_block_trace_model_path: str,
     self._thread_count = thread_count
     self._corpus_path = corpus_path
 
+    gin.parse_config(gin_config)
+    self._setup_base_policy()
+
+  def __del__(self):
+    shutil.rmtree(self._tf_base_temp_dir)
```
mtrofin, line 81: iiuc `__del__` happens at GC. Can we do this deterministically, e.g. based on a scope?

compiler_opt/es/regalloc_trace/regalloc_trace_worker.py
```diff
@@ -44,8 +46,16 @@ class RegallocTraceWorker(worker.Worker):
...
-  def __init__(self, clang_path: str, basic_block_trace_model_path: str,
-               thread_count: int, corpus_path: str):
+  def _setup_base_policy(self):
+    self._tf_base_temp_dir = tempfile.mkdtemp()
+    policy = policy_utils.create_actor_policy()
+    saver = policy_saver.PolicySaver({"policy": policy})
+    saver.save(self._tf_base_temp_dir)
+    self._tf_base_policy_path = os.path.join(self._tf_base_temp_dir, "policy")
+
+  def __init__(self, gin_config: str, clang_path: str,
```
mtrofin, line 56: can you force these to be passed by-name, I couldn't for the life of me figure out what `""` meant in tests, for the first parameter.

compiler_opt/es/regalloc_trace/regalloc_trace_worker.py
boomanaiden154: Python doesn't really have the notion of running a destructor after falling out of scope as far as I can tell, so this is sort of the closest we can get.

We could set up the runner as a context manager, but instantiating each runner with a `with` statement won't work.

compiler_opt/es/blackbox_learner.py
boomanaiden154: Numpy defaults to doubles but the rest of the code expects floats, so they're converted here just to ensure that both sides expect the same width types.

compiler_opt/es/regalloc_trace/regalloc_trace_worker.py
boomanaiden154: Done.

compiler_opt/es/regalloc_trace/regalloc_trace_worker.py
mtrofin: Right, my point is that the dtor will be run non-deterministically, at GC.

Why is `with` not an option for each worker? Because it'd add an extra responsibility?

Oh, but this is a worker-lifetime bound tempdir, correct? so it's a bit of "best effort", right?

boomanaiden154: Yeah, it is bound by the lifetime of the worker. If it leaks it isn't a big deal, but not ideal.

GIven it's in the constructor of the worker, we'd have to setup the worker as a context manager which is incompatible with how we have worker pools setup. The other approach would be to have the worker pool call some close method or something that deletes the temp dir/cleans up, but there will still be issues with that where things don't get cleaned up, so I think this is a reasonable enough approach.

mtrofin: add a comment that this is best effort, as it happens at shutdown and, if the shutdown is forced (not uncommon on cloud setups), :shrug: (and probably doesn't matter since there's more aggressive cleanup happening there anyway)

mtrofin (approved):

PR #453, opened 2025-02-27
Separate worker and worker manager args

mtrofin:
Without this, it's hard to extend the argument list for the local worker manager.


boomanaiden154 (approved):
This definitely makes things a bit more clear.

LGTM.

PR #454, opened 2025-02-27
Ignore ruff C408

mtrofin:
We make use of kwargs quite a bit, and it's sometimes necessary to pass then through, e.g. to workers from the worker manager. It looks (aesthetically, at least) cleaner to pass as `key=value` rather than `"key":value`, also, we can avoid silly typos like `"k ey"`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,7 @@
 [tool.ruff]
 line-length = 103
 lint.select = [ "C40", "C9", "E", "F", "PERF", "UP", "W", "YTT" ]
-lint.ignore = [ "E722", "E731", "F401", "PERF203" ]
+lint.ignore = [ "E722", "E731", "F401", "PERF203", "C408" ]
 lint.mccabe.max-complexity = 18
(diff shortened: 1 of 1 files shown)


mtrofin:
@cclauss are there trade-offs we'd be missing by ignoring this rule? Thanks!

boomanaiden154 (approved):

cclauss:
Type `ruff rule C408` and read the ___Why is this bad?___ section.  If the rule is slowing your progress then go ahead and ignore it.

mtrofin:
> Type `ruff rule C408` and read the _**Why is this bad?**_ section. If the rule is slowing your progress then go ahead and ignore it.
> 

Thanks - got it. The perf overhead should be fine, this particular pattern happens at setup mostly.

> this setting might be helpful… https://docs.astral.sh/ruff/settings/#lint_flake8-comprehensions_allow-dict-calls-with-keyword-arguments

Oh, even better. Doing that then. Thanks!

PR #455, opened 2025-02-27
Use standard pickle, and make pickler an option

mtrofin:
We used cloudpickle to make it easy to author tests where the
test worker is defined in the test function. The problem is that
cloudpickle will also try to serialize by value classes defined in the
`__main__` module, which leads to serialization problems when e.g. the worker class
references in its members unpickleable things.

Also, revert #452 as it is addressed by this change.


boomanaiden154 (approved):
Seems reasonable enough to me.

PR #459, opened 2025-03-01
Disable pytype pyi-error globally

boomanaiden154:
This patch disable pyi-error globally. We have a lot of manual disables for this error sprinkled throughout the codebase, to the point where we should just disable it globally given the noise that it causes.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -46,7 +46,9 @@ jobs:
         task: [Typecheck, Lint, Ruff, Yapf, Test]
         include:
           - task: Typecheck
-            cmd: pytype -j auto .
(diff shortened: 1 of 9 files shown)


mtrofin:
would that let unwanted errors - i.e. cases we expect a pyi - just pass the type checker?

boomanaiden154:
> would that let unwanted errors - i.e. cases we expect a pyi - just pass the type checker?

Yes, but I'm not sure it's practically an issue. Pytype complains about this because it needs pyi files to obtain some information it needs for type checking. They exist in the vast majority of cases, but in some (with complex imports/module provides it seems), it just errors out. Given that every single error would require upstream refactoring or fixes in pytype and we've just been disabling them, I figured a global override would be reduce (the mostly useless) noise.

PR #463, opened 2025-03-05
Compute relative reward in trace blackbox evaluator

boomanaiden154:
This patch makes the trace blackbox evaluator return a relative reward rather
than the raw reward. This makes the rewards actually meaningful, and also
prevents overflowing calculatings in blackbox_optimizer which previously
made the new model weights NaN.

### DIFF
--- a/compiler_opt/es/blackbox_evaluator.py
+++ b/compiler_opt/es/blackbox_evaluator.py
@@ -23,6 +23,7 @@ from compiler_opt.distributed.worker import FixedWorkerPool
 from compiler_opt.rl import corpus
 from compiler_opt.es import blackbox_optimizers
 from compiler_opt.distributed import buffered_scheduler
+from compiler_opt.rl import compilation_runner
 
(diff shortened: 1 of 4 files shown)


compiler_opt/es/blackbox_evaluator.py
```diff
@@ -159,3 +160,17 @@ def set_baseline(self, pool: FixedWorkerPool) -> None:
...
+
+  def get_rewards(
+      self, results: list[concurrent.futures.Future]) -> list[float | None]:
+    rewards = []
+
+    for result in results:
+      if result.exception() is not None:
+        raise result.exception()
+
+      # pylint: disable=protected-access
```
mtrofin, line 172: why not make `compilation_runner._calculate_reward` "public" - just also add a test for it (pro'lly that a 0 baseline doesn't Inf)

mtrofin (approved):
lgtm, one comment.

PR #465, opened 2025-03-09
Add support for copying corpora to local disk

boomanaiden154:
This patch adds support in regalloc_trace_worker for copying the corpus
to the local disk at initilization time. This is necessary for certain
filesystems that can only be interacted with through specific APIs,
precluding their direct use for applications like clang.

### DIFF
--- a/compiler_opt/es/regalloc_trace/regalloc_trace_worker.py
+++ b/compiler_opt/es/regalloc_trace/regalloc_trace_worker.py
@@ -38,9 +38,9 @@ from compiler_opt.rl import policy_saver
 from compiler_opt.es import policy_utils
 
 
-def _make_dirs_and_copy(old_path: str, new_path: str):
-  tf.io.gfile.makedirs(os.path.dirname(new_path))
-  tf.io.gfile.copy(old_path, new_path)
+def _make_dirs_and_copy(old_file_path: str, new_file_path: str):
+  tf.io.gfile.makedirs(os.path.dirname(new_file_path))
+  tf.io.gfile.copy(old_file_path, new_file_path)
 
 
 @gin.configurable
@@ -60,8 +60,8 @@ class RegallocTraceWorker(worker.Worker):
     saver.save(self._tf_base_temp_dir)
     self._tf_base_policy_path = os.path.join(self._tf_base_temp_dir, "policy")
 
-  def _maybe_copy_corpus(self, corpus_path: str,
-                         copy_corpus_locally_path: str | None) -> str:
+  def _copy_corpus(self, corpus_path: str,
+                   copy_corpus_locally_path: str | None) -> None:
     """Makes a local copy of the corpus if requested.
 
     This function makes a local copy of the corpus if requested by copying the
@@ -69,18 +69,11 @@ class RegallocTraceWorker(worker.Worker):
     location is specified, it simply returns the existing path.
 
     Args:
-      corpus_path: THe path to the remote corpus.
-      copy_corpus_locally: THe path to copy the corpus to, or None if no
-        copying is desired.
-
-    Returns:
-      The path to the corpus that is to be used for compilation/evaluation.
+      corpus_path: The path to the remote corpus.
+      copy_corpus_locally: The local path to copy the corpus to.
     """
-    if copy_corpus_locally_path is None:
-      return corpus_path
-
     # We use the tensorflow APIs below rather than the standard Python file
-    # APIs for compatibility with internal (to Google) filesystems.
+    # APIs for compatibility with more filesystems.
 
     if tf.io.gfile.exists(copy_corpus_locally_path):
       return copy_corpus_locally_path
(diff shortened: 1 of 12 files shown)


compiler_opt/es/regalloc_trace/regalloc_trace_worker.py
```diff
@@ -27,15 +27,22 @@
...
 import gin
+import tensorflow as tf
 
 from compiler_opt.rl import corpus
 from compiler_opt.distributed import worker
 from compiler_opt.rl import policy_saver
 from compiler_opt.es import policy_utils
 
 
+def _make_dirs_and_copy(old_path: str, new_path: str):
```
mtrofin, line 41: add `_file`. And `old_file_path` and `new_file_path`. So noone believes later this applies to dirs.

compiler_opt/es/regalloc_trace/regalloc_trace_worker.py
```diff
@@ -53,9 +60,65 @@ def _setup_base_policy(self):
...
+  def _maybe_copy_corpus(self, corpus_path: str,
+                         copy_corpus_locally_path: str | None) -> str:
+    """Makes a local copy of the corpus if requested.
+
+    This function makes a local copy of the corpus if requested by copying the
+    remote corpus to a user-specified directory. If no local corpus copy
+    location is specified, it simply returns the existing path.
+
+    Args:
+      corpus_path: THe path to the remote corpus.
```
mtrofin, line 72: `THe`

compiler_opt/es/regalloc_trace/regalloc_trace_worker.py
```diff
@@ -53,9 +60,65 @@ def _setup_base_policy(self):
...
+      copy_corpus_locally: THe path to copy the corpus to, or None if no
+        copying is desired.
+
+    Returns:
+      The path to the corpus that is to be used for compilation/evaluation.
+    """
+    if copy_corpus_locally_path is None:
+      return corpus_path
+
+    # We use the tensorflow APIs below rather than the standard Python file
```
mtrofin, line 82: Even less Google-specific, tf apis support a variety of file systems.

compiler_opt/es/regalloc_trace/regalloc_trace_worker.py
```diff
@@ -53,9 +60,65 @@ def _setup_base_policy(self):
     saver.save(self._tf_base_temp_dir)
     self._tf_base_policy_path = os.path.join(self._tf_base_temp_dir, "policy")
 
-  def __init__(self, *, gin_config: str, clang_path: str,
-               basic_block_trace_model_path: str, thread_count: int,
-               corpus_path: str):
+  def _maybe_copy_corpus(self, corpus_path: str,
```
mtrofin, line 63: should this be a corpus.py utility?

compiler_opt/es/regalloc_trace/regalloc_trace_worker.py
```diff
@@ -53,9 +60,65 @@ def _setup_base_policy(self):
...
+                         copy_corpus_locally_path: str | None) -> str:
+    """Makes a local copy of the corpus if requested.
+
+    This function makes a local copy of the corpus if requested by copying the
+    remote corpus to a user-specified directory. If no local corpus copy
+    location is specified, it simply returns the existing path.
```
mtrofin, line 69: but why? if you don't want copying, why call this function?

basically, I'd call this function `copy`, add it in corpus.py, and make both paths mandatory.

compiler_opt/es/regalloc_trace/regalloc_trace_worker_test.py
```diff
@@ -26,15 +27,16 @@
 from compiler_opt.rl import corpus
 
 
-def _setup_corpus(corpus_dir: str) -> list[corpus.ModuleSpec]:
+def _setup_corpus(corpus_dir: str,
+                  has_thinlto: bool = False) -> list[corpus.ModuleSpec]:
   modules = [
-      corpus.ModuleSpec("module_a", 1, ("-fmodule-a",), True),
-      corpus.ModuleSpec("module_b", 1, ("-fmodule-b",), True)
+      corpus.ModuleSpec("module_a.o", 1, ("-fmodule-a",), True),
```
mtrofin, line 33: why does this matter now?

compiler_opt/es/regalloc_trace/regalloc_trace_worker.py
mtrofin: also, in this scenario (trace-based etc...) do we always want to copy the whole corpus, or just specific modules?

boomanaiden154: > should this be a corpus.py utility?

I think it's a bit too specific to be generic to `corpus.py`, and no one else needs this functionality currently. If other workers in the future end up needing to copy the corpus locally, we can potentially look at generalizing it and moving it to somewhere like `corpus.py`.

> also, in this scenario (trace-based etc...) do we always want to copy the whole corpus, or just specific modules?

We want to copy the whole corpus, at least for now. The corpus that is being copied would have already been subsetted/minimized.

compiler_opt/es/regalloc_trace/regalloc_trace_worker_test.py
boomanaiden154: This is how the corpora are actually laid out and how the `_setup_corpus` utility sets them up. Now that we're actually looking at the file paths rather than just checking mocks, they need to actually exist.

compiler_opt/es/regalloc_trace/regalloc_trace_worker.py
boomanaiden154: Felt it might make things a bit cleaner. Your way I think makes the code more readable, so opted for that.

mtrofin: ok but then `copy_corpus_locally_path` is just `str` and the doc needs updating

mtrofin (approved):
lgtm, a comment left to address.

compiler_opt/es/regalloc_trace/regalloc_trace_worker.py
boomanaiden154: Good point. Fixed.

PR #466, opened 2025-03-10
[Bug Fix] Wrapping args of LocalWorkerPoolManager with worker_kwargs

svkeerthy:


boomanaiden154 (approved):
LGTM once CI is fixed.

I've been meaning to write out a base class for worker managers so we can detect these issue at type checking time, but haven't gotten around to it. Maybe this evening.

mtrofin (approved):

boomanaiden154:
> Would a type annotation have helped?

I would think so. The type checker would then have an actual `__init__` to look at rather than nothing since `worker_manager_class` is untyped and thus doesn't have anything explicit associated with it.

PR #467, opened 2025-03-10
Adding '--warning-suppression-mappings' in flags to delete

svkeerthy:


boomanaiden154 (approved):
LGTM.

With this everything goes through? Looking at my gin files, it seems like there were a couple more flags that I needed to delete:

https://github.com/google/ml-compiler-opt/blob/a12a09909df42f159d2a39e0affd1fb5e787801f/compiler_opt/es/regalloc_trace/gin_configs/blackbox_learner.gin#L31C1-L32C1

svkeerthy:
Oh really! I could see that things are working upon adding this flag.

> LGTM.
> 
> With this everything goes through? Looking at my gin files, it seems like there were a couple more flags that I needed to delete:
> 
> https://github.com/google/ml-compiler-opt/blob/a12a09909df42f159d2a39e0affd1fb5e787801f/compiler_opt/es/regalloc_trace/gin_configs/blackbox_learner.gin#L31C1-L32C1

PR #468, opened 2025-03-10
[buildbot] Adjust gemmlowp_INSTALL_LIBDIR based on CMake's choices

DataCorrupted:
Some of our servers default installation to `lib64` instead of `lib`. This patch asks CMakeCache which one did it use.


mtrofin (approved):

petrhosek:
Rather than relying on the the defaults inferred by the `GNUInstallDirs` module and reading those out of the `CMakeCache.txt` file, I think it'd be cleaner to explicitly set `CMAKE_INSTALL_BINDIR`, `CMAKE_INSTALL_LIBDIR` and `CMAKE_INSTALL_INCLUDEDIR` .

DataCorrupted:
> explicitly set `CMAKE_INSTALL_LIBDIR`

It seems other repos are already doing this, I'll just follow suit and set it to `lib` instead.

boomanaiden154 (approved):

PR #477, opened 2025-03-28
Add option to generate_default_trace to output keys

boomanaiden154:
This patch adds an option to generate_default_trace to output the keys associated with the examples. This is primarily intended for use in extracting functions that contain eviction decisions to improve the efficiency of the new training workflow for ES regalloc trace modelling. However, this might also be useful in other circumstances if more introspection into the data is needed.

### DIFF
--- a/compiler_opt/tools/generate_default_trace.py
+++ b/compiler_opt/tools/generate_default_trace.py
@@ -66,7 +66,7 @@ _GIN_BINDINGS = flags.DEFINE_multi_string(
     'Gin bindings to override the values set in the config files.')
 _KEYS_FILE = flags.DEFINE_string(
     'keys_file', None,
-    'THe path to the file to write out the keys encountered.')
+    'The path to the file to write out the keys encountered.')
 
 
 class FilteringWorker(worker.Worker):
@@ -207,8 +207,7 @@ def generate_trace(worker_manager_class: type[
 
   if _KEYS_FILE.value is not None:
     with open(_KEYS_FILE.value, 'w', encoding='utf-8') as keys_file:
-      for key in all_keys:
-        keys_file.write(f'{key}\n')
+      keys_file.write('\n'.join(str(key) for key in all_keys) + '\n')
 
 
 if __name__ == '__main__':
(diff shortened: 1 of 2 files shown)


compiler_opt/tools/generate_default_trace.py
```diff
@@ -64,6 +64,9 @@
 _GIN_BINDINGS = flags.DEFINE_multi_string(
     'gin_bindings', [],
     'Gin bindings to override the values set in the config files.')
+_KEYS_FILE = flags.DEFINE_string(
+    'keys_file', None,
+    'THe path to the file to write out the keys encountered.')
```
cclauss, line 69:
```suggestion
    'The path to the file to write out the keys encountered.')
```

compiler_opt/tools/generate_default_trace.py
```diff
@@ -196,6 +205,11 @@ def generate_trace(worker_manager_class: type[
         f'succeeded, and {total_training_examples} trainining examples '
         'written')
 
+  if _KEYS_FILE.value is not None:
+    with open(_KEYS_FILE.value, 'w', encoding='utf-8') as keys_file:
+      for key in all_keys:
+        keys_file.write(f'{key}\n')
```
cclauss, line 211: Use a single `.write()` call instead of multiple calls.
```suggestion
      keys_file.write('\n'.join(str(key) for key in all_keys) + '\n')
```

mtrofin (approved):
took me a bit to remember what "keys" refers to. These are function names for regalloc, right?

LGTM after @cclauss 's feedback is addressed.

mtrofin:
maybe say "keys (e.g. function names for regalloc)" in the commit message

compiler_opt/tools/generate_default_trace_test.py
```diff
@@ -111,9 +111,15 @@ def test_generate_trace(self, mock_get_runner):
...
                                              'output_performance'),
-    ):
+        keys_file=os.path.join(tmp_dir.full_path, 'keys_file')):
       generate_default_trace.generate_trace()
 
+    with open(
+        os.path.join(tmp_dir.full_path, 'keys_file'),
+        encoding='utf-8') as keys_file:
+      keys = [key_line.strip() for key_line in keys_file.readlines()]
+      self.assertListEqual(keys, ['default', 'default', 'default', 'default'])
```
mtrofin, line 121: can we have a test that's not just saying "default"?

boomanaiden154: I switched it up to add the pid and then just assert the start. Not sure it does that much though...

There's no easy way from what I can tell to pass a specific ID to each runner that we could then use to split the keys.

PR #479, opened 2025-03-31
Fix saving best policy in ES

boomanaiden154:
The variable name was bad and it wasn't caught by static analysis as we were setting a new variable before actually using it. This caused the best model each iteration to get saved whether or not it was actually better than all previous models.

### DIFF
--- a/compiler_opt/es/blackbox_evaluator_test.py
+++ b/compiler_opt/es/blackbox_evaluator_test.py
@@ -32,7 +32,7 @@ class BlackboxEvaluatorTests(absltest.TestCase):
         blackbox_test_utils.ESWorker,
         count=3,
         worker_args=('',),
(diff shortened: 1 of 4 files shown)


mtrofin (commented):
Is there a test that could have caught this?

boomanaiden154:
> Is there a test that could have caught this?

Not currently. I opened https://github.com/google/ml-compiler-opt/pull/480 to add a test (which still does not catch this) as it was much easier than I thought. I'll add a regression test.

boomanaiden154:
Tests added. Note that this PR now includes the changes from #480 so that it will actually work. I couldn't set it up as a stacked PR as I had already created the PR using a branch from my fork.

mtrofin (approved):

PR #484, opened 2025-04-03
Add a script to compile a corpus

boomanaiden154:
This patch adds a script that enables compiling a corpus with the
regalloc trace worker. This is intended for use when partially compiling
a corpus and enables extracting out only the functions that matter for
register allocation decisions.

This is intended to be used before training begins to run all the middle
end optimizations. This then allows for functions to be extracted after
all IPO has taken place so we can then compile them all separately
by just running the backend. This script also supports compiling all the
way to asm to enable testing with basic_block_trace_model.

This script is typically used with --mode=full so that basic blocks in
the trace that are in functions that do not have regalloc decisions can
get serialized. The script is then run again with --mode=bc to run all
middle end optimziations so that the extract_functions script can be run
on the partially compiled corpus to extract all of the functions with
regalloc decisions that we are interested in compiling during training.
--mode=asm is used for validating that an extracted corpus can be used
for modelling without having to start the training process.

### DIFF
--- a/compiler_opt/tools/regalloc_trace/compile_corpus.py
+++ b/compiler_opt/tools/regalloc_trace/compile_corpus.py
@@ -75,15 +75,17 @@ def main(_) -> None:
       replace_flags=replace_compilation_flags,
   )
 
-  logging.info("Compiling corpus.")
+  logging.info('Compiling corpus.')
   compile_start = time.time()
+  # pylint: disable=missing-kwoa
   worker = regalloc_trace_worker.RegallocTraceWorker(
       gin_config=gin.operative_config_str())
+  # pylint: disable=protected-access
   worker._build_corpus(train_corpus.module_specs, _OUTPUT_PATH.value,
                        _TFLITE_POLICY_PATH.value)
   compile_end = time.time()
   compile_duration = compile_end - compile_start
-  logging.info(f'Compilation took {compile_duration}s')
+  logging.info('Compilation took %ds', compile_duration)
 
 
 if __name__ == '__main__':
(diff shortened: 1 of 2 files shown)


mtrofin:
Can you elaborate the scenario, where and when does this get run as part of the training workflow?

boomanaiden154:
> Can you elaborate the scenario, where and when does this get run as part of the training workflow?

Added to the commit description.

mtrofin:
> > Can you elaborate the scenario, where and when does this get run as part of the training workflow?
> 
> Added to the commit description.

OK, so this means the input to trace training includes the output of _this_? (can you also clarify that :))

compiler_opt/tools/regalloc_trace/compile_corpus.py
```diff
@@ -0,0 +1,92 @@
...
+_OUTPUT_PATH = flags.DEFINE_string(
+    'output_path', None, 'The path to the output.', required=True)
+_GIN_FILES = flags.DEFINE_multi_string(
+    'gin_files', [], 'List of paths to gin configuration files to use.')
+_GIN_BINDINGS = flags.DEFINE_multi_string(
+    'gin_bindings', [], 'Gin bindings to override the values in config files.')
+_TFLITE_POLICY_PATH = flags.DEFINE_string(
+    'tflite_policy_path', None,
+    'The path to the TFLite policy to use, if there is one.')
+_MODE = flags.DEFINE_enum('mode', 'full', ['full', 'bc', 'asm'],
```
mtrofin, line 38: do you need all these for your scenario right now?

boomanaiden154: Technically we do not need the `asm` mode, but it makes testing that the extracted corpus and serialized BBs are actually able to supply all the data needed for modelling the traces much easier. `full` and `bc` are definitely needed to serialize BBs and to generate the post middle end bitcode.

boomanaiden154:
> OK, so this means the input to trace training includes the output of this? (can you also clarify that :))

Technically from the output of other tools that take the output of this script as their input. I outlined exact usage in the commit description.

mtrofin (approved):

PR #487, opened 2025-04-03
Add extract_functions script

boomanaiden154:
This just parses some flag values and passes them along to
extract_functions_lib so that the relevant functionality can actually be
invoked from the command line.


mtrofin:
elaborate the scenario - why and when is this tool useful?

mtrofin (approved):

PR #493, opened 2025-04-14
[WIP] Inlining worker for ES (DO NOT SUBMIT)

svkeerthy:


compiler_opt/es/blackbox_evaluator.py
```diff
@@ -97,12 +123,44 @@ def get_results(
...
+    if len(results) != len(self._baselines):
+      raise RuntimeError(
+          'The number of results does not match the number of baselines.')
+
+    policy_results = super().get_rewards(results)
+
+    rewards = []
+    for i in range(len(policy_results)):
+      policy_result = policy_results[i]
+      baseline = self._baselines[i]
```
cclauss, line 157: `range(len())` is almost always a sign to use `enumerate()` or `zip()` instead.
```suggestion
    for policy_result, baseline in zip(policy_results, self._baselines, strict=True):
```

PR #495, opened 2025-04-24
Make compile_corpus BC mode produce an actual output corpus

boomanaiden154:
When using the compile_corpus script to partially compile a corpus
(running middle-end optimizations), we need to make sure that the output
is actually another corpus. This entails making sure the file
suffix/extension is the same as the original corpus and that .cmd files
are copied over. This patch adds that funtionality.


mtrofin:
test?

boomanaiden154:
> test?

I added a test for the new functionality in `regalloc_trace_worker`. There isn't any testing setup in `compile_corpus` because it is theoretically simple. If it ends up getting more complex/we run into bugs, I'll make sure to add some testing infra for it, but for now I think it should be fine.

mtrofin (approved):

PR #497, opened 2025-04-24
Add rich error reporting for add/delete/replace flag overlap

boomanaiden154:
This patch gives some additional information when the add/delete/replace
flag lists overlap. This makes things a bit easier to debug without
having to deeply inspect the values or manually insert some print
statements.

### DIFF
--- a/compiler_opt/rl/corpus.py
+++ b/compiler_opt/rl/corpus.py
@@ -343,10 +343,20 @@ class Corpus:
 
     # don't use add/remove for replace
     add_keys = {k.split('=', maxsplit=1)[0] for k in additional_flags}
-    if add_keys.intersection(
-        set(replace_flags)) or set(delete_flags).intersection(
-            set(replace_flags)) or add_keys.intersection(set(delete_flags)):
-      raise ValueError('do not use add/delete flags to replace')
+    if add_keys.intersection(set(replace_flags)):
+      raise ValueError(
+          'Do not use add/delete flags to replace flags: add flags and '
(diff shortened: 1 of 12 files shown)


compiler_opt/rl/corpus.py
```diff
@@ -343,10 +343,20 @@ def __init__(self,
...
-            set(replace_flags)) or add_keys.intersection(set(delete_flags)):
-      raise ValueError('do not use add/delete flags to replace')
+    if add_keys.intersection(set(replace_flags)):
+      raise ValueError(
+          'Do not use add/delete flags to replace flags: add flags and '
+          'replace flags intersect: '
+          f'{add_keys.intersection(set(replace_flags))}')
+    elif set(delete_flags).intersection(set(replace_flags)):
+      raise ValueError(
+          'Do not use add/delete flags to replace flags: delete flags and '
```
mtrofin, line 353: is the string the exact same? factor it out, e.g

msg = 'Do not use add/delete flags to replace flags: add flags and '
          'replace flags intersect: '
          f'{details}'

and then you can `msg.format(details = <whatever it is.. .add_keys.intersection... whatever>)`

mtrofin (approved):

compiler_opt/rl/corpus.py
boomanaiden154: They're not exactly the same (what intersects differs), and I think factoring it out would just make the code quite a bit less readable. Happy to follow up post commit if opinions continue to differ though.

PR #499, opened 2025-04-30
Add hooks in the trainer

boomanaiden154:
This patch adds hooks to the train process. This enables adding
functions that can run every n steps, like a function that evaluates the
model and writes the results out to tensorboard for later evaluation.

### DIFF
--- a/compiler_opt/rl/trainer.py
+++ b/compiler_opt/rl/trainer.py
@@ -14,6 +14,7 @@
 """LLVM Policy Trainer."""
 
 import time
+from typing import Callable
 
 from absl import logging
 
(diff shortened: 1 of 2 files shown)


compiler_opt/rl/trainer.py
```diff
@@ -239,3 +244,8 @@ def train(self, dataset_iter, monitor_dict, num_iterations: int):
         self._update_metrics(experience, monitor_dict)
         self._log_experiment(loss.loss)
         self._save_checkpoint()
+
+        if hooks:
```
mtrofin, line 248: `if hooks is not None:`

mtrofin (approved):

svkeerthy (approved):
LGTM upon the suggested changes.

PR #501, opened 2025-04-30
Add shuffle_repeat_count parameter to dataset function

boomanaiden154:
This patch adds a shuffle_repeat_count parameter to tfrecord/file
datasets. This enables setting the repeat_count to something other
than None. Setting the repeat_count to None means it will repeat
indefinitely making it difficult to know when we have iterated over the
entire dataset. Setting it explicitly allows us to iterate over the
entire dataset a precise number of times when needed.

### DIFF
--- a/compiler_opt/rl/data_reader_test.py
+++ b/compiler_opt/rl/data_reader_test.py
@@ -128,6 +128,28 @@ class DataReaderTest(tf.test.TestCase, parameterized.TestCase):
     self.assertAllClose([[2.3, 2.3, 2.3], [2.3, 2.3, 2.3]], experience.reward)
     self.assertAllEqual([[1, 1, 1], [1, 1, 1]], experience.discount)
 
+  def test_create_dataset_fn_shuffle_repeat(self):
+    agent_type_override = agent_config.DQNAgentConfig
+    example = _define_sequence_example(
+        agent_type_override, is_action_discrete=True)
+
+    data_source = self._create_tfrecord_datasource(example)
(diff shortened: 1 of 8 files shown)


compiler_opt/rl/data_reader_test.py
```diff
@@ -128,6 +128,28 @@ def test_create_dataset_fn(self, test_fn, _, data_source_fn):
...
+        train_sequence_length=3,
+        shuffle_repeat_count=1)
+
+    batch_count = 0
+    for _ in dataset_fn([data_source]):
+      batch_count += 1
+
+    # Assert that we only see a fixed number of batches and only iterate over
+    # the dataset once, which in this case means we see 166 batches.
+    self.assertEqual(batch_count, 166)
```
svkeerthy, line 151: Keeping `shuffle_repeat_count` to `1` or `None` would result in the `same batch_count`, right?

boomanaiden154: Setting it to None would result in a batch_count of infinity.

svkeerthy: Ah.. Got it! Should we have a test for that too?

svkeerthy (approved):

mtrofin (approved):

compiler_opt/rl/data_reader_test.py
boomanaiden154: It's hard to test that it's actually infinite. All the other tests are also checking the `shuffle_repeat_count=None` case.

PR #502, opened 2025-05-01
Only pull unique values in generate_vocab

boomanaiden154:
This patch makes generate_vocab only consider unique values when creating quantile buckets for the individual features. There is no point in mapping multiple of the same value to different values post normalization/bucketization. This only serves to decrease dynamic range. Only looking at unique values preserves the property of increasing dynamic range around areas that have lots of slightly different values. This is especially useful when generating vocab for features that are oftentimes zero or one. In some cases I was seeing <50/1000 actual values with the rest being zeros or ones, which results in models that train poorly.


boomanaiden154:
Running some BC hparam explorations right now with one of the parameters I'm exploring being vocab. I'm ablating against the current vocab in tree, new vocab without this patch, and new vocab with this patch. I'll try and post the results here once I have them in a couple hours.

PR #503, opened 2025-05-01
Save hparams at the beginning of train_bc

boomanaiden154:
This patch moves where we save hparams from the end of the train_bc script to before we start training. This means that we can use the tensorboard hparams dashboard at the beginning of experiments and as things progress rather than only at the end when all experiments have concluded.


svkeerthy (approved):

mtrofin (approved):
🤦

svkeerthy:
Exactly what I was looking at a while back! Thanks for the patch :)

PR #506, opened 2025-05-16
Add construct_cmd_for_compilation flag to corpus

boomanaiden154:
Having this option makes things a little bit simpler when writing the
group functions library.

### DIFF
--- a/compiler_opt/rl/corpus.py
+++ b/compiler_opt/rl/corpus.py
@@ -272,7 +272,8 @@ class Corpus:
                additional_flags: tuple[str, ...] = (),
                delete_flags: tuple[str, ...] = (),
                replace_flags: dict[str, str] | None = None,
-               sampler_type: type[Sampler] = SamplerBucketRoundRobin):
+               sampler_type: type[Sampler] = SamplerBucketRoundRobin,
+               construct_cmd_for_compilation=True):
     """
     Prepares the corpus by pre-loading all the CorpusElements and preparing for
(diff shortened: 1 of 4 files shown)


compiler_opt/rl/corpus.py
```diff
@@ -272,7 +272,8 @@ def __init__(self,
                additional_flags: tuple[str, ...] = (),
                delete_flags: tuple[str, ...] = (),
                replace_flags: dict[str, str] | None = None,
-               sampler_type: type[Sampler] = SamplerBucketRoundRobin):
+               sampler_type: type[Sampler] = SamplerBucketRoundRobin,
+               construct_cmd_for_compilation=True):
```
mtrofin, line 276: document it

mtrofin: maybe "add_input_cmd_flags"

what about thinlto?

boomanaiden154: Added a docstring. I carved out ThinLTO as well. It didn't come up before because a corpus produced by `extract_functions.py` does not have any ThinLTO given they are modules produced right after the middle end is completely done.

I think `construct_cmd_for_compilation` better captures the intent than `add_input_cmd_flags`. If we set this flag to true we are adding additional flags to prepare the command line to be passed directly to a clang invocation for compilation. I think `add_input_cmd_flags` is a bit vague on what it actually does.

mtrofin (approved):

PR #507, opened 2025-05-16
Add group functions library

boomanaiden154:
This patch adds a library and associated unit tests for grouping
functions emitted from the extract_functions script for more efficient
compilation.

### DIFF
--- a/compiler_opt/tools/regalloc_trace/group_functions_lib.py
+++ b/compiler_opt/tools/regalloc_trace/group_functions_lib.py
@@ -32,7 +32,7 @@ def _get_functions_chunked_by_command_line(
     function_folder: str, delete_flags: tuple[str, ...] = ()
 ) -> dict[tuple[str], list[str]]:
   """Groups functions by their command line.
-  
+
   This function takes in a path to a corpus containing modules that all contain
   single functions (from the extract_functions script). It then arranges these
   functions by their commandline, ensuring all functions with the same
@@ -43,7 +43,7 @@ def _get_functions_chunked_by_command_line(
       to process.
     delete_flags: The flags to delete from each of the command lines. This
       should contain all flags that make each command line unique.
-  
+
   Returns:
     A dictionary mapping the command line stored as a tuple to a list of
     function names.
--- a/compiler_opt/tools/regalloc_trace/group_functions_lib.py
+++ b/compiler_opt/tools/regalloc_trace/group_functions_lib.py
@@ -30,7 +30,7 @@ class FunctionPathAndSize:
 
 def _get_functions_chunked_by_command_line(
     function_folder: str, delete_flags: tuple[str, ...] = ()
-) -> dict[str, list[FunctionPathAndSize]]:
+) -> dict[tuple[str], list[str]]:
   function_corpus = corpus.Corpus(
       data_path=function_folder,
       delete_flags=delete_flags,
@@ -65,8 +65,8 @@ def _get_functions_chunked_by_command_line(
 
 
 def _partition_functions(
-    functions_per_command_line: dict[str, list[FunctionPathAndSize]],
-    max_functions_per_chunk: int) -> dict[str, list[list[str]]]:
+    functions_per_command_line: dict[tuple[str], list[str]],
+    max_functions_per_chunk: int) -> dict[tuple[str], list[list[str]]]:
   corpus_chunks = {}
   for command_line in functions_per_command_line:
     corpus_chunks[command_line] = []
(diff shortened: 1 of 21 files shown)


compiler_opt/tools/regalloc_trace/group_functions_lib.py
```diff
@@ -0,0 +1,120 @@
...
+import dataclasses
+import os
+import math
+import subprocess
+import json
+
+from compiler_opt.rl import corpus
+
+
+@dataclasses.dataclass
```
mtrofin, line 25: freeze it?

compiler_opt/tools/regalloc_trace/group_functions_lib.py
```diff
@@ -0,0 +1,120 @@
...
+class FunctionPathAndSize:
+  path: str
+  size: int
+
+
+def _get_functions_chunked_by_command_line(
+    function_folder: str, delete_flags: tuple[str, ...] = ()
```
mtrofin, line 32: how come you only need `delete_flags` (as opposed to also add_flags, etc)?

compiler_opt/tools/regalloc_trace/group_functions_lib.py
```diff
@@ -0,0 +1,120 @@
...
+      data_path=function_folder,
+      delete_flags=delete_flags,
+      construct_cmd_for_compilation=False)
+  command_lines = {}
+
+  for module_spec in function_corpus.module_specs:
+    function_path = os.path.join(function_folder, module_spec.name + '.bc')
+    function_size = module_spec.size
+    function_path_and_size = FunctionPathAndSize(function_path, function_size)
+    module_command_line = list(module_spec.command_line)
```
mtrofin, line 44: why can't you just make it a tuple on line 44?

compiler_opt/tools/regalloc_trace/group_functions_lib.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+def _partition_functions(
+    functions_per_command_line: dict[tuple[str], list[str]],
+    max_functions_per_chunk: int) -> dict[tuple[str], list[list[str]]]:
+  corpus_chunks = {}
+  for command_line in functions_per_command_line:
+    corpus_chunks[command_line] = []
+    chunks_for_command_line = math.ceil(
+        len(functions_per_command_line[command_line]) / max_functions_per_chunk)
```
mtrofin, line 74: check and raise somewhere that max_functions_per_chunk > 0

compiler_opt/tools/regalloc_trace/group_functions_lib.py
```diff
@@ -0,0 +1,120 @@
...
+  size: int
+
+
+def _get_functions_chunked_by_command_line(
+    function_folder: str, delete_flags: tuple[str, ...] = ()
+) -> dict[tuple[str], list[str]]:
+  function_corpus = corpus.Corpus(
+      data_path=function_folder,
+      delete_flags=delete_flags,
+      construct_cmd_for_compilation=False)
```
mtrofin, line 37: used where?

compiler_opt/tools/regalloc_trace/group_functions_lib.py
```diff
@@ -0,0 +1,120 @@
...
+from compiler_opt.rl import corpus
+
+
+@dataclasses.dataclass
+class FunctionPathAndSize:
+  path: str
+  size: int
+
+
+def _get_functions_chunked_by_command_line(
```
mtrofin, line 31: add a doc comment - because a good function name is kind of hard; the function_folder contains, what, modules containing a single function?

compiler_opt/tools/regalloc_trace/group_functions_lib_test.py
```diff
@@ -0,0 +1,127 @@
...
+        corpus_folder.full_path,
+        elements=[
+            corpus.ModuleSpec(name='module1', size=5, command_line=('-cc1',)),
+            corpus.ModuleSpec(name='module2', size=5, command_line=('-cc1',))
+        ])
+    corpus_chunks = group_functions_lib.get_chunks(corpus_folder.full_path, (),
+                                                   2)
+    self.assertDictEqual(
+        corpus_chunks, {
+            ('-cc1',): [[
```
mtrofin, line 39: you're only sorting by size, and that's the same for both, so what guarantees deterministic order for the paths?

compiler_opt/tools/regalloc_trace/group_functions_lib.py
boomanaiden154: We can add/replace flags if we need to in the worker like normal. We want to delete flags here that cause the command lines to be different (so mainly anything that is a per TU output).

compiler_opt/tools/regalloc_trace/group_functions_lib.py
boomanaiden154: Good point, probably left over from some testing.

compiler_opt/tools/regalloc_trace/group_functions_lib.py
boomanaiden154: Not sure what you're commenting on here. `function_corpus` is used shortly below. We need `construct_cmd_for_compilation` to be set to `False` to avoid getting an extra `-x ir {}` stuck in the command line from the corpus tooling.

compiler_opt/tools/regalloc_trace/group_functions_lib.py
mtrofin: oh, because the add/replace would be uniformously added to all? add a comment about that

mtrofin (approved):

PR #509, opened 2025-05-16
Use unittest framework tempdirs in blackbox_learner_test

boomanaiden154:
blackbox_learner_test was using the tempdir (ie /tmp) rather than
absltest managed temp directories that are cleaned up at the end of the
test. This patch fixes it so that we are using the proper temp
directories.

Without this patch some failures for patches up the stack were not
emerging due to files that were leftover in /tmp.


mtrofin (approved):
in the title, can you replace "proper" with "unittest framework temp dir" or something like that ("proper" does not communicate anything properly <- yes I think I'm funny)

PR #510, opened 2025-05-16
Support batched bb traces in TraceBlackboxEvaluator

boomanaiden154:
This patch makes TraceBlackboxEvaluator support batched bb traces.
Essentially this means we can split all of the bb traces obtained from a
single memtrace into a bunch of chunks. We can then evaluate only one of
these chunks per perturbation. This adds some randomness to the training
which can be good and also massively speeds up evaluation.

### DIFF
--- a/compiler_opt/es/blackbox_evaluator.py
+++ b/compiler_opt/es/blackbox_evaluator.py
@@ -15,9 +15,12 @@
 
 import abc
 import concurrent.futures
+import os
+import random
 
 from absl import logging
 import gin
+import tensorflow as tf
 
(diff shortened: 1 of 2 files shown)


mtrofin (approved):
the splitting happens elsewhere though, correct? so this patch supports such a split, where the chunks are in the provided folders, right?

boomanaiden154:
> the splitting happens elsewhere though, correct? so this patch supports such a split, where the chunks are in the provided folders, right?

Yep. Splitting them beforehand should not impact efficacy of the training process and splitting them dynamically makes everything a lot more complicated dependency wise.

PR #516, opened 2025-05-25
Use flag.DEFINE_* return methods

boomanaiden154:
This enables type checking of flag variables and is the reccomended way
to consume flag values these days.

For this who have access, go/python-tips/051#use-return goes into more
detail.


svkeerthy (approved):
lgtm

PR #519, opened 2025-05-31
Add -disable-llvm-passes in compile_corpus.py

boomanaiden154:
This makes it so that we do not have to remember to add it when running training by making it correct at the time of construction. This would have prevented me from spawning a bunch of training jobs while forgetting the flag.

### DIFF
--- a/compiler_opt/tools/regalloc_trace/compile_corpus.py
+++ b/compiler_opt/tools/regalloc_trace/compile_corpus.py
@@ -14,7 +14,6 @@
 """Compiles a corpus for further downstream processing."""
 
 import os
-import shutil
 import time
 
 from absl import app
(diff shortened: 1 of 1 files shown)


compiler_opt/tools/regalloc_trace/compile_corpus.py
```diff
@@ -92,9 +89,15 @@ def main(_) -> None:
   if _MODE.value == 'bc':
     for module in train_corpus.module_specs:
       command_file = os.path.join(_CORPUS_PATH.value, module.name) + '.cmd'
-      new_command_file_path = os.path.join(_OUTPUT_PATH.value,
-                                           module.name) + '.cmd'
-      shutil.copy(command_file, new_command_file_path)
+      with open(command_file, encoding='utf-8') as command_file_handle:
+        command_tuple = tuple(command_file_handle.read().split('\0')[:-1])
+      command_tuple = command_tuple + ('-disable-llvm-passes',)
+      new_command_file_path = (
```
mtrofin, line 95: Wait, so compiling to "bc" means... what. Taking the pre-opt IR and leaving it as-is?

boomanaiden154: Taking the pre-opt IR and then running the rest of the middle end pipeline. We don't want to run any middle end passes after that when we compile to asm, so we add the -disable-llvm-passes flag to prevent them from running.

mtrofin:
Lgtm

mtrofin (approved):

PR #521, opened 2025-07-23
Restore best model test functionality

boomanaiden154:
Some of the tests around the "save best model" functionality were
changed up in 6cf15b30ea80282a7a74e7c07092af2887cc1a86 as it touched
SamplingBlackboxEvaluator which is used in the tests in a way that made
them not work. This patch restores the functionality by changing up how
the tests work a bit so they test what they are intended to.


svkeerthy (approved):
LGTM

PR #523, opened 2025-07-23
Make blackbox_learner_test use FS for seeing written policies

boomanaiden154:
This matches how the policy saver works in production. This is needed to
ensure that the test still works once we move to saving the policy in a
separate process.

### DIFF
--- a/compiler_opt/es/blackbox_learner_test.py
+++ b/compiler_opt/es/blackbox_learner_test.py
@@ -19,6 +19,7 @@ import cloudpickle
 import gin
 import numpy as np
 import numpy.typing as npt
+import pathlib
 import tensorflow as tf
 from tf_agents.networks import actor_distribution_network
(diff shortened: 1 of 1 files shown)


compiler_opt/es/blackbox_learner_test.py
```diff
@@ -115,12 +116,15 @@ def setUp(self):
                                     'policy')
     saver.save(policy_save_path)
 
-    self._saved_policies = []
+    self._iteration_policies_path = os.path.join(output_dir.full_path,
+                                                 'policies')
+    os.mkdir(self._iteration_policies_path)
```
mtrofin, line 121: `exist_ok=True`?

boomanaiden154: Added an assertion. It should not exist given it's unique per test.

PR #524, opened 2025-07-23
Save models during ES training in a separate thread

boomanaiden154:
This patch makes it so that blackbox_learner saves models during ES
training in a separate thread. There is a bit of additionaly complexity
so that we save the models concurrently while waiting for compilation
results (when the process will be blocked waiting on IO). If we just
spawn jobs whenever we run into performance problems due to the GIL and
are back where we started. We cannot use subprocesses as we cannot
pickle the policy saver function.

This extra complexity is justified through significant performance gains on
training jobs with relatively short step times (up to 20%, still single digit
percentages with step times around 5 minutes).

### DIFF
--- a/compiler_opt/es/blackbox_learner.py
+++ b/compiler_opt/es/blackbox_learner.py
@@ -17,6 +17,7 @@ from absl import logging
 import dataclasses
 import gin
 import math
+import multiprocessing
 import numpy as np
 import numpy.typing as npt
 import tensorflow as tf
@@ -162,6 +163,15 @@ class BlackboxLearner:
     self._evaluator = self._config.evaluator(self._train_corpus,
                                              self._config.estimator_type)
 
+    self._thread_pool = multiprocessing.pool.ThreadPool(processes=1)
+    self._models_to_save = []
+    self._models_to_flush = []
+
+  def __del__(self):
+    self._start_model_saving()
+    self._flush_models()
+    self._thread_pool.close()
+
   def _get_perturbations(self) -> list[npt.NDArray[np.float32]]:
     """Get perturbations for the model weights."""
     rng = np.random.default_rng(seed=self._seed)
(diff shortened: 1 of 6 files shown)


compiler_opt/es/blackbox_learner.py
```diff
@@ -225,11 +235,31 @@ def _log_tf_summary(self, rewards: list[float]) -> None:
...
-        parameters=self._model_weights, policy_name=f'iteration{self._step}')
+    self._models_to_save.append((parameters, policy_name))
+
+  def _start_model_saving(self):
+    for _ in range(len(self._models_to_save)):
+      model_parameters, model_name = self._models_to_save[0]
+      self._models_to_flush.append(
+          self._thread_pool.apply_async(self._policy_saver_fn,
+                                        (model_parameters, model_name)))
+      self._models_to_save.pop(0)
```
mtrofin, line 250: why pop and not just add each model from _models_to_save, then clear self._models_to_save?

mtrofin: same idea below

compiler_opt/es/blackbox_learner.py
```diff
@@ -225,11 +235,31 @@ def _log_tf_summary(self, rewards: list[float]) -> None:
...
+
+  def _flush_models(self):
+    for _ in range(len(self._models_to_flush)):
+      model_save_result = self._models_to_flush[0]
+      model_save_result.wait()
+      if not model_save_result.successful():
```
mtrofin, line 256: *not* successful?

mtrofin: oh you want to raise the exception? doesn't a future have an exception you could raise?

mtrofin:
> > you'd want to block further enqueuing if the queue isn't drained fast enough.
> 
> With this setup, the queue is drained every step. If it isn't drained by the time we finish getting rewards from the workers, it just waits until we are done saving the models.

I missed the `wait`. OK, makes sense. Just some nits.

PR #525, opened 2025-07-28
Integrating IR2Vec with inlining_runner

svkeerthy:
Optionally allow using IR2Vec embeddings with inlining runner. IR2Vec embeddings would be used upon specifying a valid vocabulary path.

### DIFF
--- a/compiler_opt/rl/feature_ops.py
+++ b/compiler_opt/rl/feature_ops.py
@@ -96,7 +96,6 @@ def get_ir2vec_normalize_fn(
       Setting this to True would reduce the impact of outlier embeddings and
       improve training stability, especially when embeddings from different
       batches have significantly different scales.
-
     eps: Small epsilon value added to standard deviation to prevent division
       by zero during standardization.
 
--- a/compiler_opt/rl/feature_ops_test.py
+++ b/compiler_opt/rl/feature_ops_test.py
@@ -43,35 +43,25 @@ class FeatureUtilsTest(tf.test.TestCase, parameterized.TestCase):
     self._quantile_file_dir = os.path.join(constant.BASE_DIR, 'testdata')
     super().setUp()
 
-  def _create_temp_vocab_file(self, vocab_data):
-    """Helper method to create a temporary vocab file with given data."""
-    with tempfile.NamedTemporaryFile(
-        mode='w', suffix='.json', delete=False) as f:
-      if isinstance(vocab_data, str):
-        f.write(vocab_data)  # For invalid JSON
-      else:
-        json.dump(vocab_data, f)
-      return f.name
-
   def _test_vocab_file_dimensions(self, vocab_data, expected_dimensions):
     """Helper method to test vocab file dimension reading."""
-    temp_file = self._create_temp_vocab_file(vocab_data)
-    try:
-      dimensions = feature_ops.get_ir2vec_dimensions_from_vocab_file(temp_file)
+    with tempfile.NamedTemporaryFile(
+        mode='w', suffix='.json', delete=True) as f:
+      json.dump(vocab_data, f)
+      f.flush()
+      dimensions = feature_ops.get_ir2vec_dimensions_from_vocab_file(f.name)
       self.assertEqual(expected_dimensions, dimensions)
-    finally:
-      os.unlink(temp_file)
 
   def _test_vocab_file_dimensions_with_exception(self, vocab_data,
                                                  expected_exception):
     """Helper method to test vocab file dimension reading that should raise
     exceptions."""
-    temp_file = self._create_temp_vocab_file(vocab_data)
-    try:
+    with tempfile.NamedTemporaryFile(
+        mode='w', suffix='.json', delete=True) as f:
+      json.dump(vocab_data, f)
+      f.flush()
       with self.assertRaises(expected_exception):
-        feature_ops.get_ir2vec_dimensions_from_vocab_file(temp_file)
-    finally:
-      os.unlink(temp_file)
+        feature_ops.get_ir2vec_dimensions_from_vocab_file(f.name)
 
   def test_build_quantile_map(self):
     quantile_map = feature_ops.build_quantile_map(self._quantile_file_dir)


compiler_opt/rl/feature_ops.py
```diff
@@ -80,3 +85,88 @@ def normalize(obs: types.Float):
...
+      mean = tf.math.reduce_mean(batch_embeddings, axis=0, keepdims=True)
+      std = tf.math.reduce_std(batch_embeddings, axis=0, keepdims=True)
+      standardized_embeddings = (batch_embeddings - mean) / (std + eps)
+      return standardized_embeddings
+
+    return z_score_normalize
+  # Currently, we just return the identity function for embeddings.
+  # We can extend this to include other normalizations like L2
+  # normalization if needed.
+  return lambda x: identity_fn(x, expand_dims=False)
```
boomanaiden154, line 121: Why does this need to be `identity_fn` here instead of just `tf.cast` (maybe bound with `functools.partial`)?

compiler_opt/rl/feature_ops.py
```diff
@@ -80,3 +85,88 @@ def normalize(obs: types.Float):
     return tf.concat(features, axis=-1)
 
   return normalize
+
+
+def get_ir2vec_normalize_fn(ir2vec_with_z_score_normalization: bool = False,
+                            eps: float = 1e-8):
```
boomanaiden154, line 91: Return type annotation here?

compiler_opt/rl/feature_ops.py
```diff
@@ -80,3 +85,88 @@ def normalize(obs: types.Float):
...
+
+  Args:
+    vocab_file_path: Path to the IR2Vec vocabulary JSON file.
+
+  Returns:
+    The number of dimensions in the embeddings.
+
+  Raises:
+    tf.errors.NotFoundError: If the vocabulary file cannot be found.
+    json.JSONDecodeError: If the vocabulary file contains invalid JSON.
```
boomanaiden154, line 136: I'm not sure you should list these two here. There's sort of an implict contract that the user is providing a valid JSON file.

>  You should not document exceptions that get raised if the API specified in the docstring is violated (because this would paradoxically make behavior under violation of the API part of the API).

https://google.github.io/styleguide/pyguide.html#doc-function-raises

Relevant section of the style guide.

compiler_opt/rl/feature_ops.py
```diff
@@ -80,3 +85,88 @@ def normalize(obs: types.Float):
...
+      raise ValueError('Vocabulary file embeddings must be lists')
+
+    if not first_embedding:
+      raise ValueError('Vocabulary file embeddings must be non-empty')
+
+    # Find any embedding array and return its length
+    return len(first_embedding)
+
+  except (tf.errors.NotFoundError, json.JSONDecodeError, ValueError) as e:
+    logging.error('Error reading vocab file %s: %s', vocab_file_path, e)
```
boomanaiden154, line 170: Why log and then reraise instead of just letting errors bubble up?

compiler_opt/rl/feature_ops_test.py
```diff
@@ -41,6 +43,36 @@ def setUp(self):
     self._quantile_file_dir = os.path.join(constant.BASE_DIR, 'testdata')
     super().setUp()
 
+  def _create_temp_vocab_file(self, vocab_data):
+    """Helper method to create a temporary vocab file with given data."""
+    with tempfile.NamedTemporaryFile(
+        mode='w', suffix='.json', delete=False) as f:
```
boomanaiden154, line 49: `delete=False` looks leftover from testing?

compiler_opt/rl/feature_ops_test.py
```diff
@@ -104,6 +143,65 @@ def test_normalize_fn_sqrt_z_normalization(self, with_sqrt, with_z_score,
...
+    }
+    self._test_vocab_file_dimensions(vocab_data, 5)
+
+  @parameterized.named_parameters(
+      ('empty_embedding', {
+          'section1': {
+              'instruction1': []
+          }
+      }, ValueError),
+      ('invalid_json', 'invalid json content {', json.JSONDecodeError),
```
boomanaiden154, line 178: Here and below: I'm not sure testing that `json.parse` throws a decode error on invalid JSON is useful here. Same for the file not existing.

compiler_opt/rl/feature_ops.py
svkeerthy: True! Left over after addressing some comments. Cleaned up now.

compiler_opt/rl/feature_ops.py
svkeerthy: Yeah, added! Other methods in this file do not have type annotations though!

compiler_opt/rl/feature_ops.py
```diff
@@ -80,3 +82,82 @@ def normalize(obs: types.Float):
...
+  """Return a normalization function for embeddings.
+
+  Args:
+    ir2vec_with_z_score_normalization: If True, applies z-score standardization
+      (whitening) to the embeddings per batch. This transforms the embeddings
+      to have zero mean and unit variance across the batch dimension.
+      Setting this to True would reduce the impact of outlier embeddings and
+      improve training stability, especially when embeddings from different
+      batches have significantly different scales.
+
```
boomanaiden154, line 99: Nit: No new line between arguments.

compiler_opt/rl/feature_ops_test.py
boomanaiden154: Ah, I see the problem now.

You're returning from the `with`. You should just drop the with altogether since using a context manager here doesn't make a lot of sense.

Ideally you'd create the file in a place where you can pass it around more easily before the `NamedTemporaryFile` falls out of scope.

compiler_opt/rl/feature_ops_test.py
```diff
@@ -41,6 +43,36 @@ def setUp(self):
...
+      if isinstance(vocab_data, str):
+        f.write(vocab_data)  # For invalid JSON
+      else:
+        json.dump(vocab_data, f)
+      return f.name
+
+  def _test_vocab_file_dimensions(self, vocab_data, expected_dimensions):
+    """Helper method to test vocab file dimension reading."""
+    temp_file = self._create_temp_vocab_file(vocab_data)
+    try:
```
boomanaiden154, line 59: Here and below in `_test_vocab_file_dimensions_with_exception`:

If you refactor the temp vocab file to be created in the test, you can then drop the try/finally here.

svkeerthy: You want me to create temp file locally within each test?

boomanaiden154: Just within the `_test_vocab_file_*` functions.

compiler_opt/rl/feature_ops_test.py
```diff
@@ -43,35 +43,25 @@ def setUp(self):
...
-
   def _test_vocab_file_dimensions(self, vocab_data, expected_dimensions):
     """Helper method to test vocab file dimension reading."""
-    temp_file = self._create_temp_vocab_file(vocab_data)
-    try:
-      dimensions = feature_ops.get_ir2vec_dimensions_from_vocab_file(temp_file)
+    with tempfile.NamedTemporaryFile(
+        mode='w', suffix='.json', delete=True) as f:
+      json.dump(vocab_data, f)
+      f.flush()
```
boomanaiden154, line 51: Are the flush calls just precautionary or did you actually run into problems?

Either way, we probably want to keep them because I don't think `json.dump` makes any guarantees.

boomanaiden154 (approved):
One question, otherwise LGTM.

compiler_opt/rl/feature_ops_test.py
svkeerthy: Yes, without flush we would end up with decode errors.

Issue #309, opened 2023-10-18
Rename policy_saver.PolicySaver to (e.g.) policy_saver.MLGOPolicySaver


mtrofin:
There's a tf_agents policy_saver.PolicySaver that's used in our policy_saver.PolicySaver.

boomanaiden154:
This is still open.

Harishankar14:
Thank you, will work on it.

PR #528, opened 2025-09-21
readability: alias tf_agents.policy_saver to avoid name confusion with local module

cnaples79:
## Summary
Alias  as  inside  to avoid confusion with the local module and its  class.

## Rationale
Issue #309 calls out confusion between our internal  module/class and TF‑Agents’ . The rename to  is already in place; this follow‑up alias makes the remaining references explicit that  refers to TF‑Agents.

## Changes
- Import TF‑Agents policy saver as .
- Update usages and type hints accordingly.

This is a no‑behavior change, improves clarity for readers and avoid shadowing concerns.

Refs #309


mtrofin:
Thanks - can you please rename the title with the prefix `[nfc]` (a'la LLVM. Signals to reviewers that "no tests needed")

Could you also reduce the message to e.g. what you have in the commit (pasted below for clarity)? Was the current commit message AI-generated? It's missing text, e.g. "Import TF‑Agents policy saver as .". Anyway, no need for overly-formatted stuff, the simpler message is perfect!


```
- Import tf_agents.policies.policy_saver as tf_policy_saver
- Update usages and type hints to use tf_policy_saver.PolicySaver

This clarifies that MLGOPolicySaver is our wrapper, and PolicySaver refers to TF-Agents.
```

Thanks!

PR #532, opened 2025-11-11
Profgen - tooling for extracting IR with profile info

mtrofin:
Tooling for extracting IR with profile metadata and comparing generated (by a LLM, presumably) code with an example IR with profile.

This is experimental, and there are no tests at the moment - the scripts are just orchestrating compiler tools and it would be too complicated to require those.


boomanaiden154 (approved):
Didn't do a thorough review of the code, but it seems to be in reasonable enough shape to land, especially given it's currently experimental.

PR #533, opened 2025-11-12
Profgen: import using the same pattern as in other tools

mtrofin:
we import relative to the repo root.

### DIFF
--- a/compiler_opt/tools/profgen/toolchain.py
+++ b/compiler_opt/tools/profgen/toolchain.py
@@ -48,14 +48,16 @@ class CompilationError(Exception):
 
 
 class BuildEnv:
+  """Represents a build environment, defined by the presence of LLVM tools."""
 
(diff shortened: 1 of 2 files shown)


compiler_opt/tools/profgen/toolchain.py
```diff
@@ -48,14 +48,16 @@ class CompilationError(Exception):
...
 
   def __init__(
       self,
       toolchain_path: str,
       workingdir: str,
       tool_suffix="",
-      extra_build_flags=[],
+      extra_build_flags=None,
   ):
+    extra_build_flags = [] if extra_build_flags is None else extra_build_flags
```
cclauss:
```suggestion
    extra_build_flags = extra_build_flags or []
```
All falsey values of extra_build_flags are equally dangerous as None.

PR #535, opened 2025-12-17
Pipfile.lock specific to the python version

mtrofin:
Some packages, e.g. `dm-reverb`, have different hashes between 3.10 and 3.11. This patch separates `Pipfile.lock` into per-version files, and adds a wrapper, `versioned_pipenv`, that uses the version-specific lock file. It does so by copying the version-specific `Pipfile.lock` in the repo root. Note that we gitingore that name, as it becomes a temporary artifact of running `versioned_pipenv`.

We could have made the copy elsewhere, but that makes certain variants of `pipenv` subcommands (like `run`) not work.

### DIFF
--- a/versioned_pipenv
+++ b/versioned_pipenv
@@ -4,7 +4,6 @@ import sys
 import subprocess
 import shutil
 
-
 lockfile = "Pipfile.lock"
 versioned_lockfile = f"{lockfile}.{sys.version_info.major}.{sys.version_info.minor}"
 


mtrofin:
> I'm not a big fan of a python wrapper around pipenv, and would probably prefer we just add directions to manually copy the pipfile, but I guess this works for now.

Or maybe we switch to another similar mechanism that supports this scenario (if it exists)?

PR #539, opened 2025-12-18
demo docker

mtrofin:
This is the initial step. It currently still uses the checked in gin files and it doesn't yet fix a git hash for this repo (the latter because we need at least this commit to land first). Also, pre-training (warmstart) isn't implemented yet.

see also this [RFC](https://discourse.llvm.org/t/rfc-mlgo-open-evaluation-and-new-onboarding-doc-on-llvm/88771)

Issue #540


mtrofin:
> LGTM.
> 
> I'm assuming the plan is to write a README at some point on how to run everything locally?

eventually.

> 
> Do we have performance numbers yet for how well ES training works on this dataset and what sort of performance people should expect?

No, and we don't yet have its own gin files either.

PR #555, opened 2026-04-22
Add `GradientAscentOptimizer.reset_state` to support checkpointing

mtrofin:


boomanaiden154 (commented):
I'm not sure I'm understanding the use case here.

This enables restoring a checkpoint after the optimizer has already been used, potentially with a different setup than before?

Why do we not just want to recreate the optimizer in such a case to avoid forgetting to reset some state?

mtrofin:
> I'm not sure I'm understanding the use case here.
> 
> This enables restoring a checkpoint after the optimizer has already been used, potentially with a different setup than before?
> 
> Why do we not just want to recreate the optimizer in such a case to avoid forgetting to reset some state?

good point!

PR #556, opened 2026-04-22
Passthrough for `subprocess.Popen` in `start_cancellable_process`

mtrofin:


boomanaiden154 (commented):
Is this intended exclusively for passing through `text=True` or are there other flags that are going to be used in different cases?

mtrofin:
> Is this intended exclusively for passing through `text=True` or are there other flags that are going to be used in different cases?

also cwd, piping stderr to stdout...

boomanaiden154 (approved):
LGTM.

PR #557, opened 2026-05-13
[buildbot] Install nanobind for IR2Vec Python bindings on ml-opt bots

nishant-sachdeva:
This is a prerequisite for enabling IR2Vec Python binding tests on the ml-opt buildbots (tracked in llvm/llvm-zorg [HERE](https://github.com/llvm/llvm-zorg/pull/846) )

The change itself is 
- Install `nanobind>=2.9,<3.0` on ml-opt workers at boot time.

- nanobind is a build-time dependency of LLVM's IR2Vec Python bindings, not a runtime dependency of ml-compiler-opt. Adding it to the Pipfile would misrepresent its ownership. 

- The requirement itself mirrors `llvm/tools/llvm-ir2vec/Bindings/requirements.txt` in llvm-project. ( to be merged [HERE](https://github.com/llvm/llvm-project/pull/194593))

If that constraint changes, this line should be updated accordingly. Currently, I don't see a way to have a single source of truth and propagate it to this repo.

- This PR must merge and workers must be rebooted before the llvm-zorg PR (which sets -DLLVM_IR2VEC_ENABLE_PYTHON_BINDINGS=ON) is merged.

Adding @boomanaiden154  , @svkeerthy  for reference 

- PR for llvm-project change : https://github.com/llvm/llvm-project/pull/194593 
- PR for llvm-zorg change : https://github.com/llvm/llvm-zorg/pull/846

### DIFF
--- a/buildbot/buildbot_init.sh
+++ b/buildbot/buildbot_init.sh
@@ -133,9 +133,10 @@ python3 -m pip install --break-system-packages buildbot-worker==2.9.0
 echo installed buildbot worker
 
 # Install IR2Vec Python binding dependencies.
-# Version constraints are defined in llvm-project:
+# Installs from llvm-project main; canonical source is
 # llvm/tools/llvm-ir2vec/Bindings/requirements.txt
-python3 -m pip install --break-system-packages "nanobind>=2.9,<3.0"
+python3 -m pip install --break-system-packages \
+  -r https://raw.githubusercontent.com/llvm/llvm-project/main/llvm/tools/llvm-ir2vec/Bindings/requirements.txt
 
 TF_PIP=$(sudo -u buildbot python3 -c "import tensorflow as tf; import os; print(os.path.dirname(tf.__file__))")
 
--- a/buildbot/buildbot_init.sh
+++ b/buildbot/buildbot_init.sh
@@ -132,6 +132,11 @@ popd
 python3 -m pip install --break-system-packages buildbot-worker==2.9.0
 echo installed buildbot worker
 
+# Install IR2Vec Python binding dependencies.
+# Version constraints are defined in llvm-project:
+# llvm/tools/llvm-ir2vec/Bindings/requirements.txt
+python3 -m pip install --break-system-packages "nanobind>=2.9,<3.0"
+
 TF_PIP=$(sudo -u buildbot python3 -c "import tensorflow as tf; import os; print(os.path.dirname(tf.__file__))")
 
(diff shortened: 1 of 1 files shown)


buildbot/buildbot_init.sh
```diff
@@ -132,6 +132,11 @@ popd
 python3 -m pip install --break-system-packages buildbot-worker==2.9.0
 echo installed buildbot worker
 
+# Install IR2Vec Python binding dependencies.
+# Version constraints are defined in llvm-project:
+# llvm/tools/llvm-ir2vec/Bindings/requirements.txt
+python3 -m pip install --break-system-packages "nanobind>=2.9,<3.0"
```
mtrofin: can't we just install those `requirements.txt`? (OK if you need to land this first to then land the change in LLVM)

how do we make sure the requirements on the llvm/ir2vec side don't clash with the ones from ml-compiler-opt - this is both interesting for the build bot as well as for a developer. It's OK if it's best effort, but curious if there are ways to address this.

mtrofin (approved):
lgtm but I have some questions.

buildbot/buildbot_init.sh
nishant-sachdeva: So, we're raising two things here

- We can install those `requirements.txt` . We could use a raw github URL
```
pip3 install --break-system-packages \
  -r https://raw.githubusercontent.com/llvm/llvm-project/main/llvm/tools/llvm-ir2vec/Bindings/requirements.txt
```
Here, we would have to first check this file into LLVM, and then apply this patch. But yes, we could do it this way. We should check if this kind of file-insert is alright for us

- As for any package conflicts, at the moment, we have none. IR2Vec only needs nanobind. But in the larger scheme of things, I don't think that there's a perfect answer here. 
  - We can document the deps in both requirements.txt and buildbot_init.sh and add disclaimers that these need to stay in sync 
  - and just accept that there's going to be some sync-overhead.

mtrofin: Ack re. requirements.txt - not saying we should, just curious if there was a reason not to. Right now, maybe best to leave as-is, as we would only need to untangle one extra package and its dependency issues, should they arise.

Re package sync - makes sense; probably once we unblock ourselves from TF 2.15 this becomes more manageable (I'm speculating that the probability of package conflicts is lower and lowerable when living in the present).

boomanaiden154 (approved):
LGTM.

PR #558, opened 2026-05-21
Factor permutation management in a class

mtrofin:
Besides factoring, this also addresses 2 problems:
- when pruning, in the antithetic case, we need to prune the pair. The optimizers expect that
- adjusting weights by the same distribution misses the scale of the weight. this favors, then, tuning small weights.

### DIFF
--- a/compiler_opt/es/blackbox_learner.py
+++ b/compiler_opt/es/blackbox_learner.py
@@ -25,6 +25,7 @@ from typing import Protocol
 
 from compiler_opt.distributed.worker import FixedWorkerPool
 from compiler_opt.es import blackbox_optimizers
+from compiler_opt.es import perturbations
 from compiler_opt.rl import corpus
 from compiler_opt.es import blackbox_evaluator  # pylint: disable=unused-import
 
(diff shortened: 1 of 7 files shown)


compiler_opt/es/blackbox_learner.py
```diff
@@ -173,20 +153,11 @@ def __del__(self):
...
-    """Get perturbations for the model weights."""
-    rng = np.random.default_rng(seed=self._seed)
-    return [
-        rng.normal(size=len(self._model_weights)) *
-        self._config.precision_parameter
-        for _ in range(self._config.total_num_perturbations)
-    ]
-
-  def _update_model(self, perturbations: list[npt.NDArray[np.float32]],
+  def _update_model(self, pert: list[npt.NDArray[np.float32]],
```
boomanaiden154, line 156: Is there a reason the parameter name needs to change here? I find `perturbations` much easier to read than `pert`.

compiler_opt/es/perturbations.py
```diff
@@ -0,0 +1,98 @@
...
+      seed: Seed for the random number generator.
+    """
+    self._dimension = dimension
+    self._precision_parameter = precision_parameter
+    self._perturbation_scale_vector = perturbation_scale_vector
+    self._total_num_perturbations = total_num_perturbations
+    self._is_antithetic = is_antithetic
+    self._rng = np.random.default_rng(seed=seed)
+
+  def get_next_perturbations(self) -> list[npt.NDArray[np.float32]]:
```
boomanaiden154, line 48: Nit: Are these necessarily `next`? There's no actual state here.

compiler_opt/es/blackbox_optimizers.py
```diff
@@ -483,15 +473,16 @@ def set_state(self, state: SequenceOfFloats) -> None:
...
-def normalize_function_values(
-    function_values: FloatArray,
-    current_value: float) -> tuple[FloatArray, list[float]]:
+def normalize_function_values(function_values: FloatArray,
+                              current_value: float) -> tuple[FloatArray, float]:
   values = function_values.tolist()
   values.append(current_value)
   mean = sum(values) / float(len(values))
   stdev = np.std(np.array(values))
+  stdev = stdev if stdev > 1e-8 else 1.0
```
boomanaiden154: We do we want to reset to 1 instead of clamping to some low value?

`1e-8` is already going to really blow up the normalized values.

compiler_opt/es/blackbox_learner.py
mtrofin: one of the linters complains that it redefines a `perturbations` in the outer scope.

compiler_opt/es/perturbations.py
mtrofin: but every time you call it you get something else. `next` suggests that.

compiler_opt/es/blackbox_optimizers.py
mtrofin: what's the alternative here?

boomanaiden154: Discussed offline, clamping is probably what makes the most sense here.

compiler_opt/es/perturbations.py
boomanaiden154: next to me implies state. Maybe something like `new` would work better?

boomanaiden154 (approved):
LGTM.

PR #564, opened 2026-06-16
Make `start_cancellable_process` arguments more pass-through to `Popen`

mtrofin:
This makes it easy to configure other args to `Popen`, like `env`. We still return just `stdout`, meaning that if we want `stderr` we need to pipe it. TBD if we ever have a scenario where we want them separate.


boomanaiden154 (approved):
LGTM.

PR #572, opened 2026-08-06
rl/log_reader: use NumPy view in _add_feature

Arjunmehta312:
## Summary

Use the existing zero-copy NumPy view exposed by
`LogReaderTensorValue` when populating protobuf repeated fields in
`_add_feature`.

## Motivation

`LogReaderTensorValue` already provides a NumPy view over the underlying
tensor buffer. Extending protobuf repeated fields from that view avoids
iterating through the ctypes-backed sequence while preserving identical
serialized output.

## Validation

- Serialized `SequenceExample` output is byte-for-byte identical.
- Extended `compiler_opt/rl/log_reader_test.py` with an additional serialized-output assertion.
- `ruff check` passes on the modified files.
- Benchmarks on representative synthetic workloads measured approximately a 15% improvement in `read_log_as_sequence_examples`.

### DIFF
--- a/compiler_opt/rl/log_reader.py
+++ b/compiler_opt/rl/log_reader.py
@@ -232,7 +232,7 @@ def _add_feature(se: tf.train.SequenceExample, spec: tf.TensorSpec,
     lst = f.float_list.value
   else:
     lst = f.int64_list.value
-  lst.extend(value)
+  lst.extend(value.to_numpy())
 
 
 def read_log_as_sequence_examples(
--- a/compiler_opt/rl/log_reader_test.py
+++ b/compiler_opt/rl/log_reader_test.py
@@ -275,6 +275,8 @@ feature_lists {
 }
 """, tf.train.SequenceExample())
     self.assertProtoEquals(expected_ctx_0, seq_examples['context_nr_0'])
+    self.assertEqual(seq_examples['context_nr_0'].SerializeToString(),
+                     expected_ctx_0.SerializeToString())
 
   def test_errors(self):
     logfile = self.create_tempfile()


mtrofin:
"Benchmarks on representative synthetic workloads measured approximately a 15% improvement in read_log_as_sequence_examples" what are the representative synthetic workloads?

Arjunmehta312:
The benchmark used synthetic logs in the same "simple log format" that `read_log_as_sequence_examples` consumes: a JSON header followed by raw binary tensor buffers, mirroring the feature mix used by the supported problem configs (inlining, regalloc) — one `float32` and one `int64` feature tensor per observation, plus a `float32` score tensor. Feature tensors were large (multi-million elements, multi-MB buffers) to reflect production logs.

Measurement: `read_log_as_sequence_examples` was run end-to-end on such a log, with the old and new `_add_feature` interleaved (25 samples each) and compared by median with 95% confidence intervals (disjoint). Depending on the run the improvement was ~13-17% (~15%). In every run the serialized `SequenceExample` output (21,971,863 bytes) was byte-for-byte identical between the two variants.

mtrofin:
> The benchmark used synthetic logs in the same "simple log format" that `read_log_as_sequence_examples` consumes: a JSON header followed by raw binary tensor buffers, mirroring the feature mix used by the supported problem configs (inlining, regalloc) — one `float32` and one `int64` feature tensor per observation, plus a `float32` score tensor. Feature tensors were large (multi-million elements, multi-MB buffers) to reflect production logs.
> 
> Measurement: `read_log_as_sequence_examples` was run end-to-end on such a log, with the old and new `_add_feature` interleaved (25 samples each) and compared by median with 95% confidence intervals (disjoint). Depending on the run the improvement was ~13-17% (~15%). In every run the serialized `SequenceExample` output (21,971,863 bytes) was byte-for-byte identical between the two variants.

Can you contribute the benchmark?

Arjunmehta312:
Yes, added `compiler_opt/rl/log_reader_benchmark.py` in commit 9f1ebf0.

It generates a synthetic log in the same simple log format (JSON header + raw tensor buffers, float32 and int64 feature tensors per observation, float32 score tensor), measures `read_log_as_sequence_examples` for the current `_add_feature` against a reference copy of the previous implementation in interleaved order, and reports median/mean/p95 with a 95% confidence interval. It also asserts the serialized output of the two variants is byte-for-byte identical.

Usage, e.g.:

```
python compiler_opt/rl/log_reader_benchmark.py --observations=8 --elems=2100000 --samples=10
```

The default parameters are close to the workload the ~15% figure was measured on; the size is configurable so it can be scaled down for quick checks. It is a standalone script (not collected by pytest) and does not add a CI dependency.

compiler_opt/rl/log_reader_benchmark.py
```diff
@@ -0,0 +1,157 @@
...
+import tempfile
+import time
+
+from absl import app
+from absl import flags
+import tensorflow as tf
+
+from compiler_opt.rl import log_reader
+
+flags.DEFINE_integer("observations", 8, "Number of observations to log.")
```
mtrofin: can you use the `_OBSERVATIONS = flags.DEFINE_integer(...` style see e.g. compiler_opt/tools/generate_default_trace.py

compiler_opt/rl/log_reader_benchmark.py
```diff
@@ -0,0 +1,157 @@
...
+flags.DEFINE_integer("elems", 2_100_000, "Elements per feature tensor.")
+flags.DEFINE_integer("samples", 10, "Interleaved samples per variant.")
+
+FLAGS = flags.FLAGS
+
+
+def _write_log(fname: str, observations: int, elems: int) -> None:
```
mtrofin: elem_count or num_elem, otherwise it reads like these are the actual elements

compiler_opt/rl/log_reader_benchmark.py
```diff
@@ -0,0 +1,157 @@
...
+    if spec.dtype not in log_reader._dtype_to_ctype:
+        raise ValueError(f"Unsupported dtype: f{spec.dtype}")
+    if spec.dtype in [tf.float32, tf.float64]:
+        lst = f.float_list.value
+    else:
+        lst = f.int64_list.value
+    lst.extend(value)
+
+
+def _time_parse(fname: str, add_feature) -> tuple[float, dict[str, tf.train.SequenceExample]]:
```
mtrofin: would timeit be more canonical?

compiler_opt/rl/log_reader_benchmark.py
```diff
@@ -0,0 +1,157 @@
...
+    """Times read_log_as_sequence_examples with the given _add_feature."""
+    log_reader._add_feature = add_feature  # pylint: disable=protected-access
+    start = time.perf_counter()
+    result = log_reader.read_log_as_sequence_examples(fname)
+    elapsed = time.perf_counter() - start
+    return elapsed, result
+
+
+def _stats(samples: list[float]) -> tuple[float, float, float, tuple[float, float]]:
```
mtrofin: the repo already depends on scipy, you can use that for e.g. p95, etc.

compiler_opt/rl/log_reader_benchmark.py
```diff
@@ -0,0 +1,157 @@
...
+regalloc problem configs.
+
+The end-to-end runtime of read_log_as_sequence_examples is measured for the
+current _add_feature implementation and for a reference copy of the previous
+one. The two variants are measured in interleaved order, and the serialized
+output of the two is checked to be byte-for-byte identical. Timings are
+machine-dependent; results are reported as median, mean, p95 and a 95%
+confidence interval.
+"""
+
```
mtrofin, line 35: can you add a usage example as a comment - if the default flag values are fine, perfect (i.e. the goal is to get a reader confident in how to use this "as intended". They can figure out later if they want to adjust flags)

mtrofin (commented):
can you also clarify if you used AI for this patch, and either way, simplify the commit message. It doesn't need all the flowery headings, which just pull on the attention of a reader. It can just say it makes (which) performance improvement and adds a benchmark demonstrating it, plus the numbers. I'd be surprised if you need more than 2 sentences.

compiler_opt/rl/log_reader_benchmark.py
Arjunmehta312: Done - switched to the `_OBSERVATIONS = flags.DEFINE_integer(...)` style as in compiler_opt/tools/generate_default_trace.py (commit 1d5db98).

compiler_opt/rl/log_reader_benchmark.py
Arjunmehta312: Done - renamed to `elem_count` (commit 1d5db98).

compiler_opt/rl/log_reader_benchmark.py
Arjunmehta312: Done - the interleaved measurements now use timeit (commit 1d5db98).

compiler_opt/rl/log_reader_benchmark.py
Arjunmehta312: Done - stats now use numpy/scipy (np.percentile for p95, scipy.stats.sem + t.interval for the 95% CI) in commit 1d5db98.

compiler_opt/rl/log_reader_benchmark.py
Arjunmehta312: Done - added a usage example to the module docstring; the default flag values are the intended workload (commit 1d5db98).

Arjunmehta312:
Thanks for the review.

Regarding AI: this patch was developed with the help of an LLM coding assistant (similar to Cursor); the change itself is small and was reviewed and validated before submission.

I've also simplified the message as suggested - it's now two sentences: the change makes `read_log_as_sequence_examples` ~15% faster (~13-17% across runs, medians with disjoint 95% CIs) and adds `compiler_opt/rl/log_reader_benchmark.py` to reproduce the measurement.

All benchmark comments are addressed in commit 1d5db98 (flag style, `elem_count` rename, timeit, numpy/scipy stats, usage example).

mtrofin:
see the license header error, otherwise lgtm.

Arjunmehta312:
@mtrofin
Hi please review and merge if possible 
Thanks again :)

mtrofin:
see failing checks. I think you can verify locally if you look at the ci definition (.github/workflows/main.yml) for the yapf and pylint commands, just make sure you use the yapf and the pylint from the supported env (python 10 or 11, with the packages installed by ./versioned_pipenv... like the README says)

mtrofin:
It's still failing, see the errors. 

When patching, could you please not add AI-style summary to the fixing commit, it's stating the obvious and basically amounts to noise.

